PR #5, opened 2023-09-02
Positional embedding xformers fix

sagadre:
Green is after fix
Screen Shot 2023-09-01 at 9 09 18 PM


open_lm/params.py
```diff
@@ -428,6 +428,12 @@ def parse_args(args):
         help='Replace the network linear layers from the bitsandbytes library. '
         'Allows int8 training/inference, etc.'
     )
+    parser.add_argument(
+        "--xformers-rotary",
```
mitchellnw, line 432: Can we call this flag something like "--old-xformers-rotary" so that it still makes sense when they fix the bug?

sagadre: hmm so when this gets fixed upstream this flag will use the fix so it won’t be “old” i guess we want this to always use the old implementation and will make changes accordingly

mitchellnw: hmm that's a good point..

ludwigschmidt: Just to check: can we set up things so that when the fix happens upstream, we won't need a flag in OpenLM? I think in general the fewer flags a user needs to pass in for regular use, the better.

sagadre: that’s how things are now but i think we do need to support an “old” flag here considering that we have already trained models with the head rotary

sagadre:
confirmed that this looks good still with most up to date changes
Screen Shot 2023-09-03 at 12 53 24 AM

mitchellnw:
Amazing!

PR #7, opened 2023-09-06
Exit on NaN loss

sagadre:
In cases where loss goes to nan (e.g., bad nodes or bad hparams), we would like to raise an exception, free resources and prevent other issues like saving checkpoints and optimization states that may lead to skipped training on restarts.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -606,13 +606,16 @@ def main(args):
                 tb_writer=writer,
             )
 
+        except ValueError as e:
+            # in this case stop training, log the error and exit cleanly
             if args.distributed:
                 dist.barrier()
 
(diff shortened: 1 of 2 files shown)


mitchellnw:
Sorry this is an annoying comment but it's maybe complicating things a bit to wrap train_one_epoch in a try/except. What about just returning False when encountering a NaN loss instead of raising an exception, and otherwise returning True at the end of train_one_epoch. Then you can do something like success = train_one_epoch(...); if not success: ...

Thoughts?

sagadre:
sounds good! will make the changes and merge after a successful 1B run

mitchellnw:
LGTM, feel free to merge whenever

PR #8, opened 2023-09-07
Improved dataloading with multiple sources without resampling.

GeorgiosSmyrnis:
This PR improves the deterministic dataloading when multiple data sources are included. The current implementation works by using a manifest for each data source, and then for each epoch using that manifest to pick shards to train for this particular epoch, according to the ratio defined by the source weights.

This PR is currently a WIP, with the following tasks remaining:
- [x] Include automated tests.
- [x] Ensure deterministic training.
- [x] Compare with training with resampling.
- [x] Update README.


GeorgiosSmyrnis:
Replacement vs No Replacement

Including a file for comparison between the sampling with and without replacement.

GeorgiosSmyrnis:
Following discussion with @gabrielilharco, splitting of the shards is no longer done strictly across epochs, but rather across chunks. Each chunk has roughly the number of samples required for the epoch, so if the number of tokens is large enough then there will be no difference.

This change was done in order to fix the issue of having only a few shards (less than the number of workers) within an epoch. This alleviates this issue by assigning more shards to an epoch, if needed.

open_lm/file_utils.py
```diff
@@ -93,7 +93,7 @@ def get_metadata_file(path):
     out = [json.loads(o) for o in out.decode('utf-8').split('\n')[:-1]]
     return out
 
-def get_shards_for_epoch(num_samples, epoch, path):
+def get_shards_for_chunk(num_samples, chunk, path):
```
sagadre, line 96: can we define what a chunk is here with a comment? is it a shard? a group of shards?

open_lm/file_utils.py
```diff
@@ -111,15 +111,39 @@ def get_shards_for_epoch(num_samples, epoch, path):
...
+    return True
+
+def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards_needed):
+    samples_per_source = [weights[i] * num_samples / sum(weights) for i in range(len(weights))]
+    
+    shard_strings_per_source = []
+    next_chunk = starting_chunk
+    shard_list_per_source = [[] for _ in range(len(paths))]
+    num_samples_per_source = [0 for _ in range(len(paths))]
+    while not enough_shards(shard_list_per_source, min_shards_needed):
```
sagadre, line 130: is it possible this will infinite loop? maybe add some error handling in that case? what happens in the case where epochs*num_train_tokens > number of tokens in the dataset?

open_lm/file_utils.py
GeorgiosSmyrnis: It's a group of shards.
I'll add a comment clarifying this.

open_lm/file_utils.py
GeorgiosSmyrnis: Good point - some error handling is needed here.

open_lm/main.py
```diff
@@ -143,7 +143,15 @@ def load_optimizer(args, model, optimizer, scaler):
         logging.info(f"=> WARNING: not resuming optimizer.")
 
 
-def save_checkpoint(args, model, optimizer, scaler, completed_epoch, evaluation_loss):
+def load_data_chunk(args):
```
sagadre, line 146: maybe call this function `load_checkpoint`?

open_lm/params.py
```diff
@@ -90,6 +90,8 @@ def parse_args(args):
     )
     parser.add_argument(
         "--dataset-metadata",
+        type=str,
```
sagadre, line 93: nit: dataset manifest instead of metadata?

open_lm/main.py
GeorgiosSmyrnis: I wanted to keep this separate since loading the chunk id only makes sense when loading without resampling (whereas the existing ```load_model``` handles the actual model loading. It might be clearer this way I believe.

open_lm/params.py
```diff
@@ -90,6 +90,8 @@ def parse_args(args):
...
+        nargs="+",
```
sagadre, line 94: are multiple manifest files supported (as suggested by the `nargs="+"` flag)? looking at `get_metadata_file(...)` it looks like only 1 file can be passed

GeorgiosSmyrnis: Multiple manifest files are supported - ```get_metadata_file``` is just called separately on each.

tests/test_dataset_deterministic.py
```diff
@@ -0,0 +1,122 @@
...
+from open_lm.main import random_seed
+from open_lm.data import get_wds_dataset
+from open_lm.file_utils import get_string_for_epoch, get_metadata_file, get_shards_for_chunk
+from open_lm.params import parse_args
+from pathlib import Path
+
+NUM_SAMPLES = 100000
+
+# Update this to two data sources with webdataset, each with their own manifest.
+INPUT_PATHS = [
```
sagadre, line 17: any way we can create a `/tests/assets` folder and put a small amount of mock data there for testing rather than pointing to local paths?

open_lm/main.py
sagadre: oh i see, u only load checkpoint to get the "data_chunk" field, makes sense then!

tests/test_dataset_deterministic.py
sagadre: for this also probably good if we can check in some of our val data as opposed to rpj or pile shards

GeorgiosSmyrnis: Added a ```/tests/assets``` with a few tarfiles.

GeorgiosSmyrnis: For now I just added some of our training shards, lmk if we want to change that.

open_lm/file_utils.py
GeorgiosSmyrnis: Updated with a comment on this.

open_lm/file_utils.py
GeorgiosSmyrnis: Added some additional error handling, and found a corner case.
If you run out of shards and have to loop back, it is kind of problematic with the current string notation.

I updated the notation to list all shards e.g. ```pipe aws s3 cp s3://shards/shard_{00000, 00002, 00004}.tar -```, which I think has the desired effect of allowing arbitrary shards (not just continuous ones).

open_lm/params.py
GeorgiosSmyrnis: Changed this into dataset-manifest.

GeorgiosSmyrnis:
In the above implementation, there is a corner case where if the final group of shards does not have enough samples to cover a single epoch, it is discarded. This will matter if we exhaust the dataset, so I believe it should be handled as well.

open_lm/file_utils.py
```diff
@@ -93,33 +94,88 @@ def get_metadata_file(path):
...
+
+def source_exhausted(paths, shard_list_per_source):
+    for i, source in enumerate(paths):
+        data = get_metadata_file(source)
+        if len(data) < len(shard_list_per_source[i]):
+            return True
+    return False
+
+
+def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards_needed):
```
achalddave, line 148: This is a pretty nontrivial function now, with a seemingly benign name. Could we add a docstring and/or possibly rename? It looks to me like the right docstring is:

```
"""
Collect shards to train on for this epoch, along with metadata.

This function looks through the list of all shards and returns a subset of shards
that result in >= num_samples samples and >= min_shards_needed shards. Note that
we cannot simply compute samples based on the number of shards, since shards can
have varying numbers of samples.
"""
```

and rename to `_get_shards_for_epoch`?

open_lm/file_utils.py
```diff
@@ -93,33 +94,88 @@ def get_metadata_file(path):
...
+    if weights is None:
+        weights = [1.0 / len(paths) for _ in range(len(paths))]
+    needed_samples_per_source = [int(np.ceil(weights[i] * num_samples / sum(weights))) for i in range(len(weights))]
+    shard_strings_per_source = []
+    next_chunk = starting_chunk
+    shard_list_per_source = [[] for _ in range(len(paths))]
+    num_samples_per_source = [0 for _ in range(len(paths))]
+    while not enough_shards(shard_list_per_source, min_shards_needed) or not enough_samples(num_samples_per_source, needed_samples_per_source):
```
achalddave, line 156: I'm a little confused here; why do we need a loop here if get_shards_for_chunk also has a loop that packs shards until we have >= needed_samples_per_epoch? Is it only for the the last chunk, where we may have fewer samples than needed?

If so, I would have thought we should handle this in get_shards_for_chunk, but I suppose it's possible that we have one data source with not enough samples, and another with more than enough samples, and things work out for us serendipitously. If my understanding is correct, I think no changes needed, but might be good to drop a comment to this effect.

achalddave (approved):

achalddave:
Approving because I think we've tested and iterated on this a bit, but left a few comments that would be nice-to-haves in this PR or next. I believe @Vaishaal is also planning to take a look, though, and we can wait for that before merge.

PR #12, opened 2023-09-12
Add generation scripts

achalddave:
This PR adds two generation scripts. One works with a HuggingFace wrapper (giving access to all of HF `generate`) functionality, and the other is a manual decoding script (initially from @gabrielilharco, then modified by @mitchellnw and @jmercat). I'm PR'ing both as they are useful for complementary purposes, and can help debug each other. Right now, the two scripts return the same outputs when `--temperature` is set to 0, and similar outputs (qualitatively) when temperature is not 0.

In addition, this PR fixes an important bug in our HF wrapper: We were enabling embedding weight tying by default even when our input and output embeddings were trained to be different. This PR fixes that in the first commit.

### DIFF
--- a/open_lm/utils/transformers/config.py
+++ b/open_lm/utils/transformers/config.py
@@ -13,6 +13,7 @@ class OpenLMConfig(PretrainedConfig):
         vocab_size: int = 50304,
         post_embed_norm: bool = False,
         weight_tying: bool = False,
+        model_norm: str = "default_layer_norm",
         **kwargs
     ):
         # Used by huggingface transformers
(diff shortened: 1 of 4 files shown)


open_lm/utils/transformers/config.py
```diff
@@ -13,18 +13,20 @@ def __init__(
         vocab_size: int = 50304,
         post_embed_norm: bool = False,
         weight_tying: bool = False,
+        model_norm: str = "default_layer_norm",
```
sagadre, line 16: can we delete this file and just use `open_lm/config.py`?

sagadre: also instead of the path being `open_lm/utils/transformers..` for this code, would it make sense to rename the dir `open_lm/generation` or something like that?

achalddave: > can we delete this file and just use open_lm/config.py?

Can we do this in a separate PR? open_lm/config.py is also used for eval, so this PR would then need to modify (+test) the eval code too. 

> would it make sense to rename the dir open_lm/generation or something like that?
I think since we use this for eval, probably better to leave the name for now?

PR #24, opened 2023-09-29
Consolidate eval arguments with train arguments

achalddave:
This PR fixes (some) duplication of config, arguments, and model creation code between train and eval. Tested eval with 1b/7b models on jeopardy and hellaswag.


mitchellnw:
This looks good to me but @kernelmachine @sagadre may have thoughts

sagadre:
lgtm as well! i agree with achal's (offline) comment that we may want to consider moving all arch parameters to the model config jsons, but this seems good for now 🙂

achalddave:
thanks! merging

PR #25, opened 2023-09-29
mup simple

sagadre:
* adding mup simple where just linear layer learning rates are scaled
* parameter: `--mup-base-width` to control the `lr` scaling
* parameter: `--mup-scale-wd` to allow `lr` and `wd` to remain decoupled in `AdamW`
* formatting

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -482,7 +482,7 @@ def main(args):
                         # weights are transposed so second dim is the fan-in
                         fan_in = p.shape[-1]
                         if fan_in in linear_params:
-                            # group parameters by their fain_in value
+                            # group parameters by their fan_in value
                             linear_params[fan_in].append(p)
                         else:
                             linear_params[fan_in] = [
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -470,32 +470,53 @@ def main(args):
         named_parameters = list(model.named_parameters())
         no_decay_params = []  # to be potentially used later
 
-        if args.mup_base_fan_in:
-            linear_modules = ["in_proj", "out_proj", "w12", "w3"]
-            mup_scale = args.mup_base_fan_in / model.params.dim
-            mup_lr = args.lr * mup_scale
+        if args.mup_base_width:
+            linear_modules = ["in_proj", "out_proj", "w12", "w3", "output"]
 
-            mup_wd = args.wd
-            if args.mup_scale_wd:
-                # ensure that wd is still decoupled from lr even when the lr changes
-                mup_wd /= mup_scale
-
-            # modify all linear modules
-            linear_params = []
+            linear_params = {}
             other_params = []
+
             for n, p in named_parameters:
                 if p.requires_grad:
                     if any(m in n for m in linear_modules):
-                        linear_params.append(p)
+                        # weights are transposed so second dim is the fan-in
+                        fan_in = p.shape[-1]
+                        if fan_in in linear_params:
+                            # group parameters by their fain_in value
+                            linear_params[fan_in].append(p)
+                        else:
+                            linear_params[fan_in] = [
+                                p,
+                            ]
                     else:
+                        # all other learnable parameters that are not in linear layers
                         other_params.append(p)
 
+            optim_params = [
+                {"params": no_decay_params, "lr": args.lr, "weight_decay": 0.0},
+                {"params": other_params, "lr": args.lr, "weight_decay": args.wd},
+            ]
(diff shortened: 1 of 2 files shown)


open_lm/main.py
```diff
@@ -467,17 +469,49 @@ def main(args):
...
-            [
-                {"params": no_decay_params, "weight_decay": 0.0},
-                {"params": params, "weight_decay": args.wd},
-            ],
-            lr=args.lr,
-            betas=(args.beta1, args.beta2),
-            eps=args.eps,
-        )
+        if args.mup_base_fan_in:
+            linear_modules = ["in_proj", "out_proj", "w12", "w3"]
```
mitchellnw, line 474: what about the final head?

open_lm/main.py
```diff
@@ -467,17 +469,49 @@ def main(args):
...
-        optimizer = optim.AdamW(
-            [
-                {"params": no_decay_params, "weight_decay": 0.0},
-                {"params": params, "weight_decay": args.wd},
-            ],
-            lr=args.lr,
-            betas=(args.beta1, args.beta2),
-            eps=args.eps,
-        )
+        if args.mup_base_fan_in:
```
mitchellnw, line 473: maybe this should be base width and not base fan-in, because fan-in is not fixed throughout the network (fan-in = # of inputs, so w3 has a different fan-in than in_proj, out_proj, w12)

open_lm/main.py
```diff
@@ -467,17 +469,49 @@ def main(args):
...
+            linear_modules = ["in_proj", "out_proj", "w12", "w3"]
+            mup_scale = args.mup_base_fan_in / model.params.dim
+            mup_lr = args.lr * mup_scale
+
+            mup_wd = args.wd
+            if args.mup_scale_wd:
+                # ensure that wd is still decoupled from lr even when the lr changes
+                mup_wd /= mup_scale
```
mitchellnw, line 481: nice!

open_lm/main.py
```diff
@@ -467,17 +469,49 @@ def main(args):
...
+                    if any(m in n for m in linear_modules):
+                        linear_params.append(p)
+                    else:
+                        other_params.append(p)
+
+            optimizer = optim.AdamW(
+                [
+                    {"params": no_decay_params, "lr": args.lr, "weight_decay": 0.0},
+                    {"params": other_params, "lr": args.lr, "weight_decay": args.wd},
+                    {"params": linear_params, "lr": mup_lr, "weight_decay": mup_wd},
```
mitchellnw, line 497: you'll have to special case w3 here

open_lm/main.py
sagadre: nice, yeah added `"output"` as well

open_lm/main.py
sagadre: now i am looping over all linear layers to get the specific fan-in from the weight shapes and creating param groups accordingly

open_lm/main.py
sagadre: sweet! changed!

open_lm/main.py
```diff
@@ -467,17 +469,70 @@ def main(args):
...
+                if p.requires_grad:
+                    if any(m in n for m in linear_modules):
+                        # weights are transposed so second dim is the fan-in
+                        fan_in = p.shape[-1]
+                        if fan_in in linear_params:
+                            # group parameters by their fan_in value
+                            linear_params[fan_in].append(p)
+                        else:
+                            linear_params[fan_in] = [
+                                p,
```
mitchellnw, line 489: this formatting is a bit strange?

open_lm/main.py
```diff
@@ -467,17 +469,70 @@ def main(args):
...
+                        other_params.append(p)
+
+            optim_params = [
+                {"params": no_decay_params, "lr": args.lr, "weight_decay": 0.0},
+                {"params": other_params, "lr": args.lr, "weight_decay": args.wd},
+            ]
+
+            for fan_in in linear_params:
+                # add param group for each fan_in
+                mup_scale = args.mup_base_width / fan_in
```
mitchellnw, line 502: We want to scale by base_fan_in / fan_in so maybe something like this is good
```
mup_scale = args.mup_base_width * (args.model_dim / fan_in) / fan_in
```

mitchellnw:
Just thought of another thing -- you may have to use --fsdp-use-orig-params flag if using this for FSDP as FSDP may chunk together many params if not, and then you may not be able to use different LRs.

PR #26, opened 2023-10-01
Mup

reinhardh:
This is an implementation for mu parameterization following 
[1] Table 8 in http://arxiv.org/abs/2309.16620 for Adam
The  parameterization scales starting from the parameters of an existing based model. 

If the parameter   --mup_base_dim [int] is given, the model model is modified following [1] as follows:
- To change the initialization variables of the output weights to be constant [1], we fix them to 1/mup_base_dim 
- To scale the output with 1/fan_in, we multiply the output with 1 / (fan_in / base_fan_in)
- To change the learning rates of the matrix-like parameters, apart from the embedding and output layers to be 1/fan_in, we scale them with 1 / (fan_in / base_fan_in)
For our parameterization of the transformer this only requires mup_base_dim.

If the parameter   --mup_base_n_heads [int] is given, we also do, following [1]:
- To scale the attention logits with 1/head_dim, we scale them as 1/sqrt(base_head_dim) * 1/ (head_dim/base_head_dim). 
If this parameter is not given, we do not scale the attention logits, if we keep head_dim constant this still follows mu parametrization. 

We could also add Mitchel's mup_simple (http://arxiv.org/abs/2309.14322) as an option, this would be a very simple modification, we would only change the learning rates, but not the scales in the model itself.

### DIFF
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -254,6 +254,6 @@ def create_model(args):
         weight_tying=cfg["weight_tying"],
         norm_type=get_norm_class(args),
         apply_qk_norm=args.qk_norm,
-        rotary_old=args.rotary_old
+        #rotary_old=args.rotary_old
     )
     return Transformer(args)
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -131,9 +131,10 @@ class CustomAttn(nn.Module):
 
         if self.mup_base_n_heads:
             '''
-            if mup_base_n_heads = 0, we assume head_dim in the target and base model is equal
-            mup requires to scale attention as 1/d instead of 1/sqrt(d)
-            relative scaling requires 
+            if mup_base_n_heads = 0, the parameter is not given and we do not scale.
+            If head_dim==mup_base_n_heads no scaling would be required.
+            mup requires to scale attention as 1/d instead of 1/sqrt(d), 
+            thus the relative scaling as below
             '''
             base_head_dim = self.mup_base_dim // self.mup_base_n_heads
             multiplier = 1 / (self.head_dim / base_head_dim)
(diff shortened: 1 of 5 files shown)


open_lm/model.py
```diff
@@ -125,7 +129,18 @@ def forward(self, x: torch.Tensor, is_causal=True):
...
+        if self.mup_base_n_heads:
+            '''
+            if mup_base_n_heads = 0, the parameter is not given and we do not scale.
+            If head_dim==mup_base_n_heads no scaling would be required.
+            mup requires to scale attention as 1/d instead of 1/sqrt(d), 
+            thus the relative scaling as below
+            '''
+            base_head_dim = self.mup_base_dim // self.mup_base_n_heads
+            multiplier = 1 / (self.head_dim / base_head_dim)
+            output = self.attn_fn(queries, keys, vals, is_causal=is_causal) * float(multiplier)
```
mitchellnw, line 141: I believe the scaling needs to be pre-softmax -- could be applied to the keys before `self.attn_fn`

open_lm/model.py
```diff
@@ -254,6 +278,8 @@ def create_model(args):
         weight_tying=cfg["weight_tying"],
         norm_type=get_norm_class(args),
         apply_qk_norm=args.qk_norm,
-        rotary_old=args.rotary_old
+        mup_base_dim=args.mup_base_dim,
+        mup_base_n_heads=args.mup_base_n_heads,
+        #rotary_old=args.rotary_old
```
mitchellnw, line 283: Why is rotary old removed?

mitchellnw:
Awesome! It looks like Samir also has a PR for mup (but just simple version) going right now: https://github.com/mlfoundations/open_lm/pull/25. No preference on my end but we should probably choose one for further iteration/testing.

open_lm/model.py
sagadre: also is there any way we can make minimal (ideally no) changes to the model.py file? mup is kind of a niche use case for open lm, so it might be nice to have the implementation separate from the main model. maybe we can use some mixin pattern for mup instead?

PR #28, opened 2023-10-06
Fixed up generate.py, added to readme

revbucket:
- made some minor readme bookkeeping changes
- fixed some minor bug in in-memory-evals 
- changed generate.py to mimic the setup from eval, and added instructions here to the readme

### DIFF
--- a/eval/eval_openlm_ckpt.py
+++ b/eval/eval_openlm_ckpt.py
@@ -1,7 +1,7 @@
 import argparse
 import json
 import time
-import sys
+import sys; sys.path.append('..')
 from typing import List
 import builtins as __builtin__
 from open_lm.model import create_params
--- a/eval/eval_openlm_ckpt.py
+++ b/eval/eval_openlm_ckpt.py
@@ -1,7 +1,7 @@
 import argparse
 import json
 import time
-import sys; sys.path.append('..')
+import sys
 from typing import List
 import builtins as __builtin__
 from open_lm.model import create_params
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ Here we'll go over a basic example where we start from a fresh install, download
 ## Setup
 We require python >=3.9, and a current installation of pyTorch, as well as several other packages. The full list of requirements is contained in `requirements.txt` and can be installed in your python enviornment via
 ```>>> pip install -r requirements.txt```
+Next, to access `open_lm` everywhere in your virtual environment, install it using pip (from within the top level github repo)
+```>>> pip install --editable . ```
 Some considerations:
 - We like [WandB](https://wandb.ai/) and [tensorboard](https://www.tensorflow.org/tensorboard) for logging. We specify how to use these during training below.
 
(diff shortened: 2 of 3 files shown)


eval/eval_openlm_ckpt.py
```diff
@@ -1,7 +1,7 @@
 import argparse
 import json
 import time
-import sys
+import sys; sys.path.append('..')
```
achalddave, line 4: Is this necessary? If you pip install openlm, you shouldn't need this right?

achalddave (commented):
One small comment - I think we don't need path modifications after pip install. Also, could you post the output you get from a sample generation run with a 1b model just to ensure nothing broke? No worries if you don't have access, I can test it tomorrow if so.

eval/eval_openlm_ckpt.py
mitchellnw: Yea good catch +1

ludwigschmidt: Continuing the conversation that's partially on Slack here. Maybe the following is useful, I've adopted this approach in the past when developing Python packages: https://setuptools.pypa.io/en/latest/userguide/development_mode.html

revbucket: Putting what I wrote on slack publicly:
-------------------------
Re the sys.path.append('..'): yes it's a bit sketchy. I hadn't installed open_lm via pip -- I generally tend to avoid assuming we've got the pip packages (especially when we're developing and making changes). We can always point pip to the directory where things are contained, but then it's tricky to make sure we're using the current under-development code.
----------------------------------
For an example output:
passing this command:
python generate.py \
--model open_lm_1b |
--checkpoint /raid/mgj528/checkpoints/open_lm/open_lm_1B/open_lm_1b.pt \
--rotary-old \
--input-text "Deep learning is defined as "
 I get this output:

Loading model into the right classes...
--------------------------------------------------
		 Model output:
--------------------------------------------------
Deep learning is defined as 
the use of neural networks as black boxes, i.e., without providing any explicit model architecture, e.g., as a generic approach for machine learning without domain expertise or a specific algorithm, or by making use of off-the-shelf libraries, to solve a variety of machine learning problems.

A popular type of machine learning method is deep learning, which has become a standard machine learning approach. In recent years, deep learning has been used to solve a variety of problems in natural language processing (NLP), such as question answering and sentiment analysis. In this context, a great deal of research has focused on developing deep learning models for NLP tasks \cite{Sutskever2014Convolutional,Hochreiter1997Long}.

Deep learning models can be trained with limited labeled data. However, to ensure that a trained model works well, additional data must be collected and labeled. This makes it difficult to adopt deep learning models in real-world applications
--------------------------------
And re: rotary old,

I just copied the exact same thing that happens with the eval/eval_openlm_ckpt.py code:
i.e.
the rotary-old kwarg gets added via open_lm.params.add_model_args and then model gets constructed via OpenLMConfig(create_params(args))

So I have no idea if this is correct, just that it's what the eval code does

achalddave: Okay awesome! The generation looks quite reasonable, let's merge this after removing `sys.path.append`. Agree with Ludwig re: `pip install -e .`, that's my usual strategy. We can add that to README if we want.

achalddave: Thanks for the PR + testing it!

PR #29, opened 2023-10-09
Support for fused cross entropy

achalddave:
As in https://github.com/mosaicml/llm-foundry/blob/7f7c0979f8b2c7d95ce634f29fe417464d983410/llmfoundry/models/mpt/modeling_mpt.py#L722-L727

In a couple informal tests, I didn't see significant wins on a 1B model on 8 A100s (12997 tok/s/gpu -> 13092 t/s/g), but adding so we have support and can test on larger models later for speed improvements.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -29,6 +29,7 @@ from torch.distributed.fsdp import (
     ShardingStrategy,
 )
 from torch.distributed.fsdp.wrap import transformer_auto_wrap_policy
+from flash_attn.losses.cross_entropy import CrossEntropyLoss as FusedCrossEntropyLoss
 
 from .model import Block
 from .losses import CrossEntropyLossWithZLoss
@@ -566,8 +567,12 @@ def main(args):
 
         return
 
-    loss = torch.nn.CrossEntropyLoss()
+    if args.fused_xent:
+        loss = FusedCrossEntropyLoss()
+    else:
+        loss = torch.nn.CrossEntropyLoss()
     if args.z_loss_coefficient != 0.0:
+        assert not args.fused_xent
         if is_master(args):
             logging.info("Using CrossEntropyLossWithZLoss.")
         loss = CrossEntropyLossWithZLoss(args.z_loss_coefficient)
--- a/open_lm/params.py
+++ b/open_lm/params.py
@@ -166,6 +166,9 @@ def parse_args(args):
         "--warmup", type=int, default=10000, help="Number of steps to warmup for."
     )
     parser.add_argument(
+        "--fused-xent", action="store_true", default=False, help="Whether to use fused cross entropy"
+    )
+    parser.add_argument(
         "--z-loss-coefficient",
         type=float,
         default=0.0,


open_lm/model.py
```diff
@@ -142,10 +143,19 @@ def __init__(self, layer_id, args: Params):
...
 
-        self.feed_forward = xops.SwiGLU(args.dim, hidden_dim, args.dim, bias=False)
+        if args.ffn_type == "swiglu":
+            # this follows llama / lit llama -- go to multiple of 256
+            hidden_dim = 256 * ((int(2 * 4 * args.dim / 3) + 256 - 1) // 256)
+            self.feed_forward = xops.SwiGLU(args.dim, hidden_dim, args.dim, bias=False)
+        elif args.ffn_type == "gelu":
+            # Follows mosaic mpt7b
+            hidden_dim = args.dim * 4
+            self._ff_w1 = nn.Linear(args.dim, hidden_dim, bias=True)
```
sagadre, line 154: do we need the bias here?

open_lm/norms.py
```diff
@@ -69,6 +69,41 @@ def extra_repr(self) -> str:
         )
 
 
+class LPLayerNorm(torch.nn.LayerNorm):
```
sagadre, line 72: if you extend the `LayerNorm` class above on L12 which is basically `torch.nn.LayerNorm` we can also expdriment with LPLayerNorm without biases

sagadre:
maybe attach a training curve / eval perplexity for some small scale experiments (e.g., 87m params, 1.7B tokens), just to make sure there is no degradation in performance?

open_lm/model.py
achalddave: not sure, https://github.com/mosaicml/llm-foundry/blob/db2233e5c660c2c8b2b9973b4de910add1c93428/llmfoundry/models/layers/ffn.py#L27-L46 has it on by default, not sure if we want it on or not

open_lm/norms.py
achalddave: ok done! (btw is there a reason we don't subclass LayerNorm from torch.nn.LayerNorm?)

open_lm/model.py
```diff
@@ -157,17 +167,29 @@ def __init__(self, layer_id, args: Params):
...
+            )
+            # scale init by depth as in https://arxiv.org/abs/1908.11365 -- worked slightly better.
+            std = 1.0 / math.sqrt(hidden_dim)
+            std = std / math.sqrt(2 * (layer_id + 1))
+            torch.nn.init.trunc_normal_(
+                self.feed_forward.w3.weight, std=std, a=-3 * std, b=3 * std
+            )
+        elif args.ffn_type == "gelu":
+            # Copied from mosaic-ml default init.
+            torch.nn.init.kaiming_normal_(self._ff_w1.weight, a=0.0, mode="fan_in", nonlinearity="relu")
```
achalddave, line 184: @mitchellnw moving from offline discussion - do you want this to be trunc_normal_?

mitchellnw: Yea I think we should change this initialization to be the same as we do do above in swiglu. It's already fairly similar but we should probably just stick to one for consistency.

In general I think we should move away from 1/block_idx extra variance scaling to 1/total_depth extra variance scaling for all output layers -- that's what's recommended by depth-mup. Or at least we should ablate on this. My guess is that it doesn't make a very large difference.

mitchellnw: Concretely in code I would recommend we leave the biases as is but then do the following instead of Kaiming normal:

```
            std = 1.0 / math.sqrt(args.dim)
            torch.nn.init.trunc_normal_(
                self._ff_w1.weight, std=std, a=-3 * std, b=3 * std
            )
            std = 1.0 / math.sqrt(hidden_dim)
            std = std / math.sqrt(2 * (layer_id + 1))
            torch.nn.init.trunc_normal_(
                self._ff_w2.weight, std=std, a=-3 * std, b=3 * std
            )
```
and the changes I'm curious in ablating over would be
A) changing `layer_id + 1` to `args.num_layers`.
B) removing the line `std = std / math.sqrt(2 * (layer_id + 1))`

open_lm/model.py
mitchellnw: I vote to get rid of bias

achalddave:
Some results on the 160m param scale. At least low precision layer norm and fused cross-entropy have basically no effect on train loss / val ppl, while giving us a good speedup. Unclear if gelu and float16 reduce are affecting convergence, we will test at 1b next.

image

open_lm/model.py
achalddave: okay let me push this change before we kick off the 1b tests then

achalddave: ok done

open_lm/model.py
achalddave: ok done

PR #32, opened 2023-10-13
Remove fused cross-entropy

achalddave:
Upon further testing, fused cross entropy only adds a ~1% speedup, but requires adding a dependency on flash_attn and fails for certain param/batch size combinations (e.g., fails at 1B with bs=32 with inscrutable errors). For simplicity, we've decided not to use it given the minor improvements.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -26,10 +26,8 @@ from torch.distributed.fsdp import (
     FullStateDictConfig,
     StateDictType,
     CPUOffload,
-    ShardingStrategy,
 )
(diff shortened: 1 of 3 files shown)


open_lm/main.py
```diff
@@ -26,10 +26,8 @@
     FullStateDictConfig,
     StateDictType,
     CPUOffload,
-    ShardingStrategy,
```
achalddave, line 29: (we're importing this twice, removing it in this PR because I saw it and was an easy fix)

PR #33, opened 2023-10-13
[DON'T MERGE] Support llama-2

yuhui-zh15:
Step 1: `python convert_llama.py /pasteur/u/yuhuiz/data/LLAMA2/llama-2-7b/consolidated.00.pth /pasteur/u/yuhuiz/data/LLAMA2/llama-2-7b/consolidated.00.converted.pth`

Step 2: verify correctness by running `verify_converted_llama.ipynb`

Note: change `/pasteur/u/yuhuiz/data/LLAMA2/llama-2-7b/consolidated.00.pth` etc to the real path in your system.

**⚠️⚠️⚠️ DON'T MERGE! JUST FOR DEMONSTRATION PURPOSE ABOUT HOW TO USE LLAMA-2 IN OPEN_LM.**


mitchellnw:
Thanks a lot this looks really great!!! Thoughts on modifying this so that it is merge-able? Could be an even more useful resource!

PR #40, opened 2023-10-18
Fused RMSNorm

sagadre:
Adding triton kernel for rms norm and testing


achalddave:
Next step is to test this on multinode, I'll look into this after Nov 17

PR #49, opened 2023-10-20
Multinode Working Tokenize shuffle

Vaishaal:
still not as fast as I'd like, somehow combining: tokenize + cut to context operation (10 minutes) and shuffle operation (10 minutes) makes things much slower. But the current script works. Will look into speeding this up in follow up PR.


mitchellnw:
Since this doesn't effect other code feel free to merge whenever it's ready

PR #50, opened 2023-10-20
Adding unites for grad_accum (with and without fsdp)

jfisher52:
Created a unittest for testing grad accumulation with and without fsdp. Found in `tests/test_grad_accum.py` with support `shared.py`.


requirements.txt
```diff
@@ -14,3 +14,4 @@ llm-foundry
 apache_beam
 datasets
 huggingface_hub
+flash_attn
```
sagadre, line 17: i think we can remove this dependency now

tests/shared.py
```diff
@@ -0,0 +1,45 @@
+class MockArgs:
+    def __init__(self, model):
+        self.model = model #part of model config
+        self.model_norm = "gain_only_layer_norm"
+        self.rotary_old = False
+        self.qk_norm = False
+        self.train_data = ["/gscratch/cse/jrfish/open_lm/tests/shard_00000000.tar",]
```
sagadre, line 7: instead of pointing to local path can we store this is huggingface or something and then pull it automatically if someone runs unit tests?

sagadre: here is some sample data that could be used for testing: https://huggingface.co/datasets/mlfoundations/open_lm_example_data/blob/main/validation_data/shard_00000000.tar

tests/test_grad_accum.py
```diff
@@ -0,0 +1,149 @@
...
+    args.train_num_samples = args.batch_size
+    
+    # increase learning rate and remove warmup for maximize change to model weights 
+    args.lr = 2
+    args.warmup = 0
+    
+    # create models
+    random_seed()
+    model_accum_grad = create_model(args).to(args.device)
+    model_no_accum_grad = copy.deepcopy(model_accum_grad).to(args.device) #should I be using copy.deepcopy or clone? or model2.load_state_dict(model1.state_dict())
```
sagadre, line 40: nit: looks like the assert statement passes so maybe remove the comment on this line

tests/test_grad_accum.py
```diff
@@ -0,0 +1,149 @@
...
```
sagadre, line 103: nit: same thing with removing comment

tests/test_grad_accum.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+        
+    # check that models weights are similar (within some threshold)
+        for p1, p2 in zip(model_accum_grad.parameters(), model_no_accum_grad.parameters()):
+            assert torch.allclose(p1, p2, atol=threshold)
+        
+
+def test_grad_acc_fsdp(accum_freq = 2, threshold = 1e-7):
+    args = MockArgs("open_lm_11m")
+    args.fsdp = True
```
sagadre, line 91: I think we want to test with: `args.fsdp = True` and `args.fsdp_amp = True`. Also this function basically same as the other one? so maybe factor out the common stuff?

tests/test_train.py
```diff
@@ -0,0 +1,68 @@
...
+
+from open_lm.train import train_one_epoch
+from open_lm.main import random_seed
+from open_lm.model import create_model
+from open_lm.data import get_data
+from open_lm.scheduler import cosine_lr
+from tests.shared import MockArgs
+
+
+def test_grad_acc():
```
sagadre, line 12: not seeing any `assert` statements here? maybe remove this file?

tests/shared.py
```diff
@@ -0,0 +1,45 @@
+class MockArgs:
+    def __init__(self, model):
+        self.model = model #part of model config
```
sagadre, line 3: can we call `download(...)` here maybe and then directly set `self.train_data = ["./tests/shard_00000000.tar", ]`?

sagadre: or maybe `["./tests/assets/shard_00000000.tar", ]`

tests/test_grad_accum.py
```diff
@@ -0,0 +1,126 @@
...
+from open_lm.train import train_one_epoch
+from open_lm.main import random_seed
+from open_lm.model import create_model
+from open_lm.data import get_data
+from open_lm.scheduler import cosine_lr
+from tests.testing_utils import download
+from tests.shared import MockArgs
+
+
+class TestsForAccumulatingGradients:
```
sagadre, line 14: does this run the tests you do `PYTHONPATH=. pytest tests/` from the root?

tests/test_grad_accum.py
```diff
@@ -0,0 +1,126 @@
...
+    """A class of unittest to accumulating gradients (with and without fsdp)"""
+
+    # Setup data, optimizer, and other basic settings
+    args = MockArgs("open_lm_11m")
+
+    # download small shard file from huggingface
+    train_data = download("testing_data", "open_lm/open_lm/tests/")
+    args.train_data = ["open_lm/open_lm/tests/testing_data",]
```
sagadre, line 22: `"./tests/testing_data"`?

tests/test_grad_accum.py
jfisher52: I am, but it is running really slow... and not completing even though the functions are working. Can you take a look when you get a chance?

PR #54, opened 2023-10-23
Update hf_config and hf_model to support HELM eval

ruixin31:
- Update `hf_config` and `hf_model` so that there is a way to export configs without any attribute of the final `params` being serializable. 
- Ensure compatibility with existing methods


mitchellnw:
What about using https://github.com/mlfoundations/open_lm/pull/20

ruixin31:
Don't think this mixin will work with our checkpoints as it requires a little bit of additional processing, and it doesn't work with `transformers.AutoModelForCausalLM` which is what HELM is using for huggingface models.

kohpangwei:
It seems like the discussion with Niels concluded 2 weeks ago. Are we waiting on anything to merge this?

Vaishaal (approved):
this LGTM! This seems like a generally useful thing

PR #39, opened 2023-10-18
LLaMA2 weight loading

jmercat:
This is a different version of issue #36 that I have been working on in parallel. Sorry for the redundancy, I'll let the OpenLM team choose what they want to take.

Usage: 
Get LLaMA weights
Convert them using `python open_lm/utils/llama2openlm.py source_weights target_weights`
`cd eval`
`torchrun --nproc_per_node 3 python eval_openlm_ckpt.py --checkpoint ../checkpoints/llama2_7b.pt --model llama2_7b.json --eval-yaml in_memory_hf_eval.yaml --tokenizer <path_to_tokenizer>`

Here are the Jeopardy results from it:

~~metrics/jeopardy/10-shot/american_history/InContextLearningLMAccuracy: 0.5690072774887085
metrics/jeopardy/10-shot/literature/InContextLearningLMAccuracy: 0.6673469543457031
metrics/jeopardy/10-shot/science/InContextLearningLMAccuracy: 0.39915966987609863
metrics/jeopardy/10-shot/word_origins/InContextLearningLMAccuracy: 0.33424657583236694
metrics/jeopardy/10-shot/world_history/InContextLearningLMAccuracy: 0.6595174074172974~~

~~Average : 0.526~~

Edit: The previous results had a bug (the normalization layer parameter was not read from the config file but from argparse), new results are slightly better:

metrics/jeopardy/10-shot/american_history/InContextLearningLMAccuracy: 0.5738498568534851
metrics/jeopardy/10-shot/literature/InContextLearningLMAccuracy: 0.6693877577781677
metrics/jeopardy/10-shot/science/InContextLearningLMAccuracy: 0.4075630307197571
metrics/jeopardy/10-shot/word_origins/InContextLearningLMAccuracy: 0.34520548582077026
metrics/jeopardy/10-shot/world_history/InContextLearningLMAccuracy: 0.6568364500999451

Average: 0.530568516254425

Here is what I did:
- ~~Added script for weight conversion from llama to OpenLM (from https://gist.github.com/Vaishaal/b5367e9c0b6e5d59d9a7be2a4ba606a9)~~
- Added a model config for llama 7b
- Added a llama_rotary positional encoding (slightly modified from https://github.com/facebookresearch/llama/blob/main/llama/model.py to make it compatible with torch.jit)
- Modified eval_openlm_ckpt.py to accept a tokenizer input argument and load the weights with strict=False
[...]

Issue #52, opened 2023-10-22
llama2 unit tests


sagadre:
currently there is an [ipython notebook](https://github.com/mlfoundations/open_lm/blob/main/open_lm/utils/verify_converted_llama.ipynb) to check if llama2 weight conversion to the open_lm format is gucci. would be great to move this to a more formal `pytest` unit test in the `tests/` directory

PR #59, opened 2023-10-27
Load LLaMA2 from HF checkpoint

sedrick-keh-tri:
This PR allows loading from HF checkpoint on-the-fly instead of requiring the checkpoint to be pre-downloaded and saved. 

```bash
python open_lm/utils/transformers/convert_llama.py --model meta-llama/Llama-2-7b-hf --output-dir llama-7b
```
It's mostly just renaming the layers to match the conventions used in HF. One key difference is the `transform_qk`, which reorders rows of the Q and K matrices. The Q and K matrices used in HF seem to be have a different order for their rows (not sure what causes this).

I am able to replicate the Jeopardy results in #39 

A few other notes:
- changed the file to use argparse for added flexibility
- moved the file inside the transformers folder rather than outside. Otherwise, any command like `from transformers import ...` will fail


sedrick-keh-tri:
Added tests/test_llama_loading.py to verify LLaMA loading. Could possibly address #52

(Note: This test might take a few minutes to run, as compared to the other test which take a few seconds, so might want to add an option or something to exclude in `pytest tests/`)

jmercat:
> Added tests/test_llama_loading.py to verify LLaMA loading. Could possibly address #52
> 
> (Note: This test might take a few minutes to run, as compared to the other test which take a few seconds, so might want to add an option or something to exclude in `pytest tests/`)

Maybe we can handle it with a pytest slow marker (see https://github.com/mlfoundations/open_lm/pull/74#issuecomment-1808871253)

PR #62, opened 2023-10-30
Use s3 cp for file load

### LINKED ISSUE #43: Speed up loading remote checkpoints
https://github.com/mlfoundations/open_lm/blob/9ca70429b5bdf0b6ca80cfe1d619d1422e60326f/open_lm/file_utils.py#L70

fsspec is somehow really slow at loading large files in my experience, and right now we have every process reading from s3. This is quite slow at large model sizes; it would be nice to speed this up, probably via `subprocess.run("aws s3 cp ...")` in local_rank=0 and then loading locally from each worker.

achalddave:
For a 1b model (~5.7GiB), using aws s3 cp takes ~140s vs. ~530s for fsspec. Also tested boto3 (client.get_object), which took about the same time as fsspec, unfortunately.

Also cleaned up some import orders while I was at it.

Closes #43 (though there may be more speedup wins to be had!)


sagadre:
slightly concerned here as the `make_2048.py` script had an `s3 cp` like this and it led to malformed tars (from what we could tell). probably worth checking that everything getting sent

achalddave:
I think the returncode check + strict state dict checking should prevent any silent errors. I'd love a version that doesn't rely on aws s3 cp, but the alternatives seem to be significantly slower (cc @Vaishaal if you know a less hacky but still fast download option),

ludwigschmidt:
Does the discrepancy between `aws s3 cp` and `client.get_object` show up for individual objects or for copying multiple objects (e.g., matched via a `*` pattern)?

achalddave:
Individual object, unfortunately (though probably exists for multiple objects, too, but in that case, we could look into parallelizing copies etc.)

ludwigschmidt:
OK thanks! I'm not sure if client.get_object does multi-part downloads by
default. Maybe worth looking into eventually:
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3.html#using-the-transfer-manager

PR #66, opened 2023-11-03
ray tokenize: support tar file input format and pre-tokenized data

iejMac:
PR for adding support for tar file input to ray preprocessing. Plan is:

- add dataset_type arg which supports jsonl/tar
- use content_key to select sample type from the tar files
- handle correct loading of data types (not sure if I should just outsource this to some automatic decoder like wds
- optional tokenizing


iejMac:
local runs work for dataset_type="tar"; content_key="txt"

iejMac:
I think this is ready for reviews/thoughts

cc: @Vaishaal @GeorgiosSmyrnis

datapreprocess/ray/tokenize_shuffle.py
```diff
@@ -375,7 +406,7 @@ def human_to_bytes(s):
         ray.init(args.ray_address, runtime_env=runtime_env)
     num_nodes = len(ray.nodes())
     # TODO  support multiple inputs
-    input_paths = glob_files(args.input, suffix=".jsonl")
+    input_paths = glob_files(args.input, suffix=f".{args.dataset_type}")
```
GeorgiosSmyrnis, line 409: I think it would be good to split dataset_type and suffix - @jeffreywpli has told me that it is convenient to glob e.g. particular jsonls based on suffix.

GeorgiosSmyrnis (commented):
Quick question, just for me to understand something here - is the assumption that the tarfiles contain non-tokenized data in "txt" or "json" fields or not? Asking because e.g. the current version of the tokenization script outputs tarfiles that have json fields, but those contain token ids, not actual text.

iejMac:
@GeorgiosSmyrnis hmmm yeah good point. So the goal of the PR was to allow tarfile input but currently it assumes that if the key you are interested in is txt or json then it needs to be tokenized. Maybe instead the decision to tokenize should be made by `args.tokenizer != None` or something like that. So the following changes would need to happen:
- If tokenizer is passed in then tarfile[key] is interpreted as a string and is tokenized, rest happens normally
- If tokenizer is not passed in then tarfile[key] is interpreted as a pre-tokenized list.

The issue with this is it can get annoying to load these "lists" from different formats. Like if the key is numpy then you need np.load, if its json then something else etc.

datapreprocess/ray/tokenize_shuffle.py
iejMac: I see... maybe there's a way to *just* use "suffix"? Can we make the point that a tar dataset is one with `"tar" in args.suffix` or is this messy?

GeorgiosSmyrnis: That might be better - maybe something like `dataset_type = args.suffix.split(".")[-1]`?
I think it's reasonable to assume that the extension is correct.

GeorgiosSmyrnis:
I think it doesn't need any changes in how json is handled at least - if the input is tokenized it should be interpreted as `int` and if not as `str`. I might be mistaken here though, and I agree that an extra parameter is not a 100% clean solution.

I think this is a matter of what our convention is, moving on - do we allow already tokenized data in files with json/txt extensions or not?

We could also have 2 conventions, based on if the tarfiles are to be used for training or not, but I think this is messy as well.

iejMac:
yeah, to be fair, this doesn't change how json is handled it just adds more options for input format inside the tar files. Current code basically says:
- if your input tarfile has numpy arrays its pre-tokenized data and just needs to be loaded into lists
- if your input tarfile has txt, its not pre-tokenized and needs to be tokenized
- if your input tarfile has json, some key is a str and needs to be tokenized

Last two are things I'm not currently doing, just thought it might be useful. I can take them out in case we are not sure of that part yet and only keep in the tarfile of pre-tokenized numpy arrays processing code. Would that be good?

GeorgiosSmyrnis:
I think that would be good!

I think @Vaishaal made updates to the tokenization code in #85 , which change how this file works significantly - maybe best to merge that first and then reincorporate these changes?

Issue #67, opened 2023-11-03
Check if batch size changed when resuming.


GeorgiosSmyrnis:
Proposed by @achalddave, opening as a different issue to keep separate.

GeorgiosSmyrnis:
Closing this for now as the dataloader does not rely on this to work.

PR #68, opened 2023-11-03
Fix sample/step counting in the case of uneven epochs.

### LINKED ISSUE #65: Wrong token count when using --accurate-total-tokens.
Looking into the logs for runs with the --accurate-total-tokens option, the number of tokens seen reported at the end of training is smaller from the desired one, by about 250-300M.

Without the option, dataloading without replacement works properly. As such, runs that don't exhaust the available data are fine.

Tagging @sagadre @achalddave @Vaishaal. I'll look into what causes this.

GeorgiosSmyrnis:
This fixes #65 - step counting during training was implicitly assuming that all epochs had the same number of steps, which might not hold when using `--no-skip-tokens` and `--accurate-total-tokens`.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -678,6 +678,10 @@ def main(args):
                         for i in range(len(args.train_data_mix_weights))
                     ]
                     chosen_num_samples = remaining_samples_per_source
+                    logging.info(
+                        "Model has seen the desired number of tokens. Running one final epoch."
+                    )
+                    final_epoch = True
                 else:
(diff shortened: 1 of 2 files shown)


GeorgiosSmyrnis:
Keeping `samples_seen` in the checkpoint for now, and will make a second PR to change this, since #67 will also need to be addressed.

GeorgiosSmyrnis:
Merging this since the token count seems correct.

PR #69, opened 2023-11-03
Integrate manifest creation

GeorgiosSmyrnis:
Incorporate manifest creation in the tokenization script, and also makes improvements in how local files are handled.

### DIFF
--- a/datapreprocess/ray/tokenize_shuffle.py
+++ b/datapreprocess/ray/tokenize_shuffle.py
@@ -4,6 +4,7 @@ import hashlib
 import io
 import json
 import os
+import glob
 import resource
 import tarfile
 import time
@@ -36,8 +37,6 @@ import pyarrow.json
 
 from tqdm import tqdm
 from io import BytesIO
-
-from Path
 
 
 class SpecialTokens(Enum):
@@ -103,7 +102,7 @@ def parse_s3_path(s3_path):
 
 def dl_parse_s3(data, creds=None):
     worker_id = ray.get_runtime_context().get_worker_id()
-    if creds is not None:
+    if creds is not None and len(creds) > 0:
         client = boto3.client(
             "s3",
             aws_access_key_id=creds["AWS_ACCESS_KEY_ID"],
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -15,4 +15,6 @@ pysimdjson
 cloudpathlib
 apache_beam
 datasets
-huggingface_hub
\ No newline at end of file
+huggingface_hub
+loguru
```
sagadre, line 19: do we want a seperate requirements of preprocessing or a unified one? i dont have a super strong prefrence here

datapreprocess/ray/tokenize_shuffle.py
```diff
@@ -4,10 +4,12 @@
 import io
 import json
 import os
+import glob
 import resource
 import tarfile
 import time
 import traceback
+import simdjson
```
sagadre, line 12: need to be added to the requirements.txt?

datapreprocess/ray/tokenize_shuffle.py
```diff
@@ -111,10 +113,16 @@ def dl_parse_s3(data, creds=None):
         client = boto3.client("s3")
     out_dicts = []
     for path in data["path"]:
-        bucket, key = parse_s3_path(path)
-        json_lines = download_to_memory_with_progress(bucket, key).decode().splitlines()
-        jsons = [json.loads(x) for x in json_lines]
-        out_dicts += jsons
+        # Download from s3 if s3 path
+        if path.startswith("s3://"):
```
sagadre, line 117: do we wanna use fsspec here?

requirements.txt
GeorgiosSmyrnis: I'm of the opinion that a combined requirements is best here - `huggingface_hub` seems like something we want anyway.

datapreprocess/ray/tokenize_shuffle.py
GeorgiosSmyrnis: Thanks! Added.

datapreprocess/ray/tokenize_shuffle.py
GeorgiosSmyrnis: I kept this as is since there is a custom download function with progress bar in the s3 case. Not particularly inclined to either case here, unless @Vaishaal has points to make re: performance.

GeorgiosSmyrnis:
Closing this PR since the `tokenize_shuffle.py` file will be refactored.

PR #74, opened 2023-11-08
Implement kv cache

jmercat:
### Explanation:

KV cache is an inference trick to avoid re-computing the whole sequence when only one token was added. It stores keys and queries of each attention layer in a list for later use.
If `use_cache` is set to `True` HuggingFace generator automatically handles this and will only input the last token along with the KV cache when generating a sequence.
The past keys and values are extracted from the cache and concatenated to the next token in each attention layer. Only one query  attends this sequence to generate the next token.
Positional embeddings need to be offset because it can no longer rely on the position of the query in the sequence (there is only 1 query).

### Tests:

LLaMA2 and a custom trained model have been tested qualitatively with the `generation.py` script with this change (with and without the flag --use-cache). Jeopardy with LLaMA2 was also ran to ensure no regression.

### Changes:

- The model outputs past_key_values regardless of the `use_cache` (returns None if not used) thus it now returns 3 output values instead of 2 
- Optional inputs `past_key_values` and `use_cache` are taken by the model and passed all the way to the attention computation
- Positional embedding takes ~two~ an extra offset argument to offset the queries and keys in the time sequence. This is useful to offset the new query and key to the end of the kv cache sequence 
- head rotary is not compatible with offsets but will run anyway since it doesn't need to offset anything


open_lm/model.py
```diff
@@ -131,24 +131,38 @@ def __init__(self, layer_id, args: Params):
...
+        queries, keys, vals = self.pos_embed(queries, keys, vals, offset=past_length)
+
+        if past_key_value is not None and use_cache:
+            keys = torch.cat([past_key_value[0], keys], dim=1)
+            vals = torch.cat([past_key_value[1], vals], dim=1)
+        if use_cache:
+            past_key_value = (keys, vals)
+
+        # If queries and keys are the same shape, we can use causal attention
+        # Otherwise, queries are expected to be the most recent tokens, and keys are the full sequence so we use full attention
```
achalddave, line 155: Should we instead expect the caller to pass is_causal=False in this case, instead of overriding is_causal here?

open_lm/positional_embedding/rotary.py
```diff
@@ -41,42 +45,43 @@ class RotaryEmbedding(torch.nn.Module):
...
         self._cos_cached = None
         self._sin_cached = None
+        self._seq_len_cached = 0
+        self._update_cos_sin_tables(seq_len)
 
-    def _update_cos_sin_tables(self, x, seq_dimension=1):
-        seq_len = x.shape[seq_dimension]
-
+    def _update_cos_sin_tables(self, seq_len: int = None, device: torch.device = None, dtype: torch.dtype = None):
         # Reset the tables if the sequence length has changed,
```
achalddave, line 60: Is this comment outdated now?

achalddave: Also, might be worth leaving a comment explaining when seq_len can be `> self.seq_len_cached`. I assume this happens when we're generating as the sequence gets longer, but worth dropping a comment imo.

open_lm/train.py
```diff
@@ -133,7 +133,7 @@ def train_one_epoch(model, data, loss, epoch, optimizer, scaler, scheduler, args
         if args.accum_freq == 1:
             with autocast():
                 inputs, targets = sample_chunk(texts, args.seq_len, args.target_mask_left)
-                out, _ = model(inputs)
+                out, _, _ = model(inputs)
```
achalddave, line 136: We might want to change our model output to be a class so we don't have to make changes like these too often. But for now this is fine.

open_lm/model.py
jmercat: I don’t think so, the model is still causal, it’s just that when only one query is passed to generate the next token it needs to attend the whole past. It is not clear what the attention function should do when there is a different number of keys and queries with is_causal=True but with the one we have it is not what we want.

open_lm/positional_embedding/rotary.py
jmercat: Good catch and fair point, I’ll update that.
You mean explaining the assert statement right?

open_lm/train.py
jmercat: Actually we could align that class with the one already used by HF. I think it would be better, should it be part of this PR?

open_lm/model.py
achalddave: I guess I would have it error when is_causal=True and only one query is passed, but maybe that's too purist. I'm fine leaving this as is unless anyone else reading this has a strong opinion (leave a comment!)

open_lm/positional_embedding/rotary.py
achalddave: hm, I'm not seeing the assert - I meant the two if statements:
`        if seq_len is None or seq_len < self._seq_len_cached:`
and
`        if seq_len > self._seq_len_cached or self._cos_cached.device != device or self._cos_cached.dtype != dtype:`

open_lm/train.py
achalddave: yeah agreed, but let's leave it for a later PR

achalddave:
Other than the one comment about adding comments to _update_cos_sin_tables, this looks good to me! Would it be possible to document the speedup we get with KV-cache, compared to say HF LLaMa2? Doesn't need to be super thorough, just a quick gut check that we have roughly the right speedup.

achalddave:
oh and finally, could you paste the outputs of pytest so we can merge? (I believe dataloading tests currently fail due to a missing assets/ folder that we're working on adding, but the rest should pass)

jmercat:
> oh and finally, could you paste the outputs of pytest so we can merge? (I believe dataloading tests currently fail due to a missing assets/ folder that we're working on adding, but the rest should pass)

image

There were bugs but most didn't come from my changes:

- I added ignore_parse_errors to MockDataArgs.
- In train.py l.144 sample chunk I updated to the new args input
- in test accumulation I updated the SimpleModel to output 3 items (this was due to my changes)

I added a test for kv cache but it is slow so I also added a slow marker in pyproject.toml, slow tests can be marked with `@pytest.mark.slow`
Tests marked as slow can be skipped with `pytest . -m "not slow"` this might help with https://github.com/mlfoundations/open_lm/pull/59#issuecomment-1786029169

jmercat:
image

jmercat:
With the attached script for a 1b model I get the following results:

Context length | Generation length | Time with cache | Time without cache | Time gain (%)
--------------------------------------------------------------------------------------------
512            | 512               | 18.1522          | 30.4054             | 40.30          
512            | 1024              | 20.8708          | 80.5402             | 74.09          
512            | 1536              | 29.9303          | 152.6310            | 80.39          
1024           | 512               | 15.6641          | 50.1081             | 68.74          
1024           | 1024              | 25.5443          | 123.8644            | 79.38          
1536           | 512               | 19.4633          | 73.8159             | 73.63          
--------------------------------------------------------------------------------------------

[time_generate.txt](https://github.com/mlfoundations/open_lm/files/13343021/time_generate.txt)

jmercat:
I added a test that checks the generated sequences. It is not very stable:
If we generate long sequences (sometimes as little as 64 tokens), they start to diverge between cached and not cached but not between two non-cached generations. I considered that this was not a bug but the result of compounding errors so I set the test to only compare the first 32 generated tokens. This ends-up passing.

open_lm/model.py
ruixin31: I think the way they [do it](https://github.com/huggingface/transformers/blob/ac5d4cf6de24b4f7fa92996e92d1d71dd5411a6a/src/transformers/models/llama/modeling_llama.py#L886-L888) is by generating an attention mask at the out-most layer and then pass it in? This is consistent with what the [original llama](https://github.com/facebookresearch/llama/blob/main/llama/model.py#L474-L489) is doing in their repo.

ruixin31: So we could also do something like this? and remove all the `is_causal` arguments but put `attention_mask` all over the place

jmercat:
I've added some support for kv_cache  + an input sequence (prior it only worked with a unique new input). 
However, it assumes the input sequence to come directly after the kv_cache in the sequence and does not support input position indices.

I've made 2 separate tests: one for speed on a randomly initialized model. One for concistency that loads the 1B pre-trained OpenLM weights and test that the generated results are the same with kv_cache and without.

I didn't produce any test for beam search...

Two tests were not passing but I don't think it's from my changes:
image

jmercat:
Ok so now check are passing but they do not include the two tests that I added (they are marked both as slow and gpu). One of them involves downloading the 1B model and using it to generate short sequences.

open_lm/model.py
jmercat: this was solved by pushing that to the lower-level function where it makes more sense to handle it. It is now doing two things: if q sequence size is 1 but k and v are larger then it sets causal to false assuming that q is the next token (otherwise it is matched with the first kv).
If q sequence length is larger than 1 but smaller than kv then it is assumed to match the last part of the kv for the same length (instead of the first part as simply setting is_causal to true would do).

achalddave:
@jmercat thanks so much for the work on this. I had two last follow ups:
1. When we talked in person, you mentioned there was a decision about what to do when query has sequence length > 1. Could you document what your decision was in the code as a comment?
2. Is it possible to add a test with a tiny model that can run in CPU? E.g., 1-2 layers.

open_lm/model.py
```diff
@@ -76,19 +76,61 @@ class Params:
...
     mask = None
-    if is_causal:
+
+    # If queries have shape [batch, 1, heads, dim] it means there is only one query in the sequence.
+    # In this case, there is no notion of causal masking, so we can just set the mask to None.
+    # This is actually needed to get the desired behavior with seq_len=1.
+    if is_causal and queries.shape[1] == keys.shape[1]:
         mask = xops.LowerTriangularMask()
+    elif is_causal and queries.shape[1] > 1:
+        # Build causal mask that assumes queries are in the end of the sequence.
```
jmercat, line 98: @achalddave You asked to comment the assumption to consider that queries are at the end of the sequence. Did you mean somewhere else than this or more details, or this is enough?

jmercat:
I had to force a minimum version of multiprocess which depends on a newer version of dill than what apache-beam wants. I could not find a common ground so I remove apache-beam from the requirements and had to add ipython.

achalddave (approved):
This looks good to me, except as discussed just revert the formatting changes from tests/. We can add those formatting changes in a separate PR.

ruixin31:
Looks good to me except for the two things we discussed offline. I was also wondering if we could look at the coverage of tests but that could be done in a separate PR.

achalddave:
nice! For future reference and documentation, could you comment what the two things you discussed offline are, @ruixin31?

jmercat:
> nice! For future reference and documentation, could you comment what the two things you discussed offline are, @ruixin31?

So what Rui suggested is to use xformers new masking  [xops.fmha.attn_bias.LowerTriangularFromBottomRightMask()](https://facebookresearch.github.io/xformers/components/ops.html#xformers.ops.fmha.attn_bias.LowerTriangularFromBottomRightMask)
It works nicely and should be easier to read and faster when using beam search. Sadly it is not compatible with llm_foundry because their dependencies don't match (`pip install -r requirements.txt` would fail but it could be installed in 2 steps successfully...)

I added comments about that in the code but reverted to using custom masks to make installation faster (we don't use beam search now so it doesn't justify the extra step in installation)

PR #76, opened 2023-11-08
pre-commit hook

sagadre:


achalddave:
this looks good but the diff is showing up as pretty big - maybe because of the black format to 120 length? Can you reset to main and just add the precommit file? I'm worried the large-ish diff contains something else accidentally

achalddave (approved):

PR #77, opened 2023-11-08
data.py: only x%vocab_size when x < 0

iejMac:
reason for this is - if you're trying out data tokenized with a higher vocab size than you expect this will silently fail and make the data wrong


sagadre:
lgtm! would be good to get @Vaishaal's input too

sagadre (approved):

PR #80, opened 2023-11-09
Fill-in-the-middle training

sedrick-keh-tri:
Implements fill-in-the-middle training according to the [OpenAI FIM paper](https://arxiv.org/abs/2207.14255) and based on BigCode's Megatron FIM [implementation](https://github.com/bigcode-project/Megatron-LM/blob/bd0aaba3492b441d7f186bb1159fc21e1dcd7a72/megatron/data/gpt_dataset.py#L684)

I've tested this and verified that it behaves as expected. Next step would be to use it to train a model and see how it performs.

I also implemented model.resize_token_embeddings, which I initially thought we would need since we're adding new special tokens for FIM, but seems like it's not necessary since our model vocab > token vocab, so we have room to spare for new tokens.

Usage: Simply supply a value for `--fim_rate` (default is 0).

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -28,6 +28,7 @@ from torch.distributed.fsdp import (
     CPUOffload,
 )
 from torch.distributed.fsdp.wrap import transformer_auto_wrap_policy
+from transformers import AutoTokenizer
 
 from .data import proc_token
 from .model import Block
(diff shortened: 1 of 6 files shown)


achalddave:
Discussed offline with @sedrick-keh-tri  - we will move this into dataset generation instead.

PR #81, opened 2023-11-10
Fix off by one in dataloading

sagadre:


tests/shared.py
```diff
@@ -78,3 +78,6 @@ def __init__(self):
         self.world_size = 1
         self.workers = 2
         self.dataset_metadata = None
```
achalddave, line 80: I guess we should remove this now (can do in a later PR)

achalddave (approved):

PR #87, opened 2023-11-14
README update re: dataloader.

GeorgiosSmyrnis:
A more detailed description of the behavior of the dataloader without replacement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,12 @@ This will create a file called ```manifest.jsonl``` under ```/preproc_data```. T
  --logs path/to/logging/dir/
 ```
 
+### Large Scale Runs
+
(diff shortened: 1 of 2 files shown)


achalddave:
This looks good, but can we either:
(1) error if the requested number of checkpoints is higher than possible, with a warning that says what the user should do, or
(2) print a large warning message indicating that we are checkpointing less?

GeorgiosSmyrnis:
I added a warning message if this is detected, and pointing out how to alleviate it. I'm not sure if this is something that should error necessarily.

achalddave:
I assume this is out of date and we don't need this anymore right? @GeorgiosSmyrnis

GeorgiosSmyrnis:
You're right @achalddave - closing this.

PR #88, opened 2023-11-14
Fix S3 streaming issues

rams16592:
File streaming used to fail without retries so trying to load the complete file before using it downline.
Tried this with Stack data and had issues so fixed it and now streaming issues are resolved.

### DIFF
--- a/open_lm/datapreprocess/ray/tokenize_shuffle.py
+++ b/open_lm/datapreprocess/ray/tokenize_shuffle.py
@@ -154,7 +154,7 @@ def get_raw_filetype(key: str):
 
 def preprocess(
     key: str,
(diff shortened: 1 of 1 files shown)


Vaishaal:
Can we make this a flag, I think this might cause memory usage to increase?

rams16592:
@Vaishaal 
I have added a flag to enable BytesIO or BinaryIO and its defaulted to BinaryIO.

Vaishaal:
Can you make this PR with main instead?

rams16592:
@Vaishaal 
I have updated the PR to go to main

rams16592:
Updating through a new PR: https://github.com/mlfoundations/open_lm/pull/174

PR #99, opened 2023-11-19
tests: run in GitHub Actions

iejMac:
cc: @achalddave

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,53 @@
+name: Continuous integration
+
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install
+        run: |
+          sudo apt-get update
+          python3 -m venv .env
+          source .env/bin/activate
+          python -m pip install -U pip
+          make install-dev
+      - name: Lint
+        run: |
+          source .env/bin/activate
+          make lint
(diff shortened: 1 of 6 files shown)


iejMac:
what config did we run black with previously? I'm getting this:
Screenshot 2023-11-19 at 4 48 20 AM

iejMac:
gtg now but seems we might have issues running some tests on actions - https://github.com/iejMac/open_lm/actions/runs/6920555691/job/18825172957

iejMac:
idk maybe we add a flag in each test specifying if GPU is required for the test and this in combination with torch.cuda.is_available() determines if the test is run

achalddave:
This is excellent, thank you @iejMac! 

> maybe we add a flag in each test specifying if GPU is required
That's a great idea. Another implementation is we can add a marker (e.g. pytest.marker.slow like https://github.com/mlfoundations/open_lm/pull/74/files#diff-c6de978711d26a475be05d36d51cc1c1e097c35cc6cfaef47c83d553c0a533cbR73) and just skip those in the Github action.

I think this is a reasonable short term fix while we also figure out how to get a GPU for testing. I'll look into this this week.

Re: black - my guess is someone forgot to run black, or their code ran it with the wrong config. Could you run black in this commit with the right config and get that test to pass?

iejMac:
@achalddave Its ready to go. If you could:
- do a final review
- check https://github.com/iejMac/open_lm/actions/runs/6936947839/job/18870060633 to see that it all passes
then let me know if you think its good. After that I need to make one more commit changing so it runs on main and PRs to main only (right now its set to this branch so I can see tests)

iejMac:
in the future tests that require GPUs should get marked the `@pytest.mark.gpu` marker and can be run via `make test-gpu`

achalddave:
This looks great, thanks again!

achalddave (approved):

PR #100, opened 2023-11-19
main.py: log number of model parameters

iejMac:
maybe I missed if this is logged somewhere else, sorry in that case.

Harmless (I think) to log and comes in useful when playing with configs


sagadre (approved):
lgtm

PR #101, opened 2023-11-21
Limited cpu support

achalddave:
Allows `test_grad_accum` to pass when run on CPU, thereby allowing it to run in Github actions.


sagadre (approved):
lgtm

PR #102, opened 2023-11-21
Track global step manually

achalddave:
Optimizers keep implicit track of steps, but they don't provide a public API to access the step, which can cause unexpected issues. For example, our current way of accessing the step fails if we pass `use_orig_params` to FSDP.

Tested with unit tests, and also tried training+resuming an 11M model with a few steps. See screenshot showing loss + tokens continuing correctly after resume.
Screenshot 2023-11-21 at 9 54 32 AM


GeorgiosSmyrnis:
Looks good to me, thanks!

GeorgiosSmyrnis (approved):

PR #103, opened 2023-11-21
Add synthetic dataset, test for simple training with synthetic data

achalddave:
This allows us to do simple tests that call through main(), similar to open_clip: https://github.com/mlfoundations/open_clip/blob/main/tests/test_training_simple.py

### DIFF
--- a/open_lm/model_configs/open_lm_test_tiny.json
+++ b/open_lm/model_configs/open_lm_test_tiny.json
@@ -1,0 +1,9 @@
+{
+    "hidden_dim": 96,
+    "n_layers": 2,
+    "n_heads": 2,
+    "seq_len": 2048,
+    "vocab_size": 50432,
+    "post_embed_norm": false,
+    "weight_tying": false
+}
--- a/open_lm/model_configs/open_lm_test_tiny.json
+++ b/open_lm/model_configs/open_lm_test_tiny.json
@@ -1,9 +1,9 @@
 {
-    "hidden_dim": 96,
+    "hidden_dim": 16,
     "n_layers": 2,
     "n_heads": 2,
-    "seq_len": 2048,
-    "vocab_size": 50432,
+    "seq_len": 16,
+    "vocab_size": 16,
     "post_embed_norm": false,
     "weight_tying": false
 }
--- a/tests/test_training_simple.py
+++ b/tests/test_training_simple.py
@@ -1,0 +1,16 @@
+from open_lm.main import main
+
+
+def test_train_simple():
+    seq_len = 2048
+    num_batches = 5
+    batch_size = 1
+    # fmt: off
+    main([
+        "--train-num-samples", str(num_batches * seq_len),
+        "--batch-size", str(batch_size),
+        "--dataset-type", "synthetic",
+        "--model", "open_lm_11m",
+        "--epochs", "1",
+    ])
+    # fmt: on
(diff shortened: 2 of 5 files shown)


mitchellnw:
LGTM

mitchellnw (approved):

tests/test_training_simple.py
```diff
@@ -0,0 +1,16 @@
...
+    batch_size = 1
+    # fmt: off
+    main([
+        "--train-num-samples", str(num_batches * seq_len),
+        "--batch-size", str(batch_size),
+        "--dataset-type", "synthetic",
+        "--model", "open_lm_11m",
+        "--epochs", "1",
+    ])
+    # fmt: on
```
sagadre, line 16: probably need something like:
`success = main(main_args)`
`assert success != -1`

achalddave: main() should throw an exception if anything breaks, which will cause the test to fail. This is probably ideal for  tests, running from command line, and for development (so you don't need to think about returning 0/1 from main and just throw an exception instead). What do you think?

sagadre: agreed, though right now main returns -1 in many failure cases. so we can either add throwing to this PR or check for -1?

achalddave: oh, didn't even realize we did that. I'll add throwing to this PR.

achalddave: added! changed all `return -1` to `raise ValueError` instead

sagadre: nice! lgtm!

tests/test_training_simple.py
```diff
@@ -0,0 +1,16 @@
...
+def test_train_simple():
+    seq_len = 2048
+    num_batches = 5
+    batch_size = 1
+    # fmt: off
+    main([
+        "--train-num-samples", str(num_batches * seq_len),
+        "--batch-size", str(batch_size),
+        "--dataset-type", "synthetic",
+        "--model", "open_lm_11m",
```
iejMac, line 13: maybe we should also add a "test config" which is even smaller? Like width is fine but maybe let's make it a 2-3 layer transformer? Call it `open_lm_test_config.json` or sth

achalddave: good idea, added

open_lm/model_configs/open_lm_test_tiny.json
```diff
@@ -0,0 +1,9 @@
+{
+    "hidden_dim": 96,
+    "n_layers": 2,
+    "n_heads": 2,
+    "seq_len": 2048,
+    "vocab_size": 50432,
```
iejMac, line 6: being a bit tedious here so feel free to ignore but couldn't we go down to some other setting for testing - like character level tokenization (would make vocab_size much smaller) and shorter ctx len?

Right now even with hidden dim 96 you will have at least `96 * 50432 = 4841472` embedding params and this last matmul for the output probs will be slow

iejMac: this is obv fine for now if it works

achalddave: not tedious at all, makes sense, especially for this test. made everything smaller

iejMac:
LGTM! Unit tests pass in <3min which is super nice

PR #105, opened 2023-11-22
Support json.gz.

GeorgiosSmyrnis:
Decoding was done outside of webdataset, which was not allowing for json.gz items to be loaded.

This PR moves decoding for json files inside the webdataset, and allows for `args.data_key == "json.gz"`.


sagadre (approved):
nice, lgtm. can we get a version bump tho for downstream?

GeorgiosSmyrnis:
Added version bump and merged.

PR #106, opened 2023-11-22
fixed start index in sample_chunk

jmercat:
faulty line: `start_idx = torch.randint(0, chunk.shape[1] - args.seq_len + 1, (1,)).item()`
if start_idx = chunk.shape[1] - args.seq_len + 1, then:
`targets = chunk[:, start_idx + 1 : start_idx + args.seq_len + 1]` becomes:
```
targets = chunk[:, chunk.shape[1] - args.seq_len + 2 :chunk.shape[1] - args.seq_len + 1 + args.seq_len + 1]
            = chunk[:, chunk.shape[1] - args.seq_len + 2 :chunk.shape[1] + 2]
            = chunk[:, chunk.shape[1] - args.seq_len + 2 :chunk.shape[1]]
```
thus targets sequence shape is < input sequence shape and the loss raises an error.

### DIFF
--- a/open_lm/train.py
+++ b/open_lm/train.py
@@ -88,7 +88,7 @@ def sample_chunk(chunk, args):
     if chunk.shape[1] == args.seq_len + 1:
         start_idx = 0
     elif chunk.shape[1] > args.seq_len + 1:
-        start_idx = torch.randint(0, chunk.shape[1] - args.seq_len + 1, (1,)).item()
+        start_idx = torch.randint(0, chunk.shape[1] - args.seq_len, (1,)).item()
     else:
         raise Exception(f"Invalid sequence length: Sequence length {args.seq_len} > {chunk.shape[1]} Chunk size")
 


iejMac:
Yes, good catch! You're right but torch.randint is exclusive on the higher value so instead it should just be:
```
start_idx = torch.randint(0, chunk.shape[1] - args.seq_len, (1,)).item()
```

So then the highest value possible is `start_idx = chunk.shape[1] - args.seq_len - 1` meaning:
```
targets = chunk[:, start_idx + 1 : start_idx + args.seq_len + 1]
targets = chunk[:, chunk.shape[1] - args.seq_len - 1 + 1 : chunk.shape[1] - args.seq_len - 1 + args.seq_len + 1]
targets = chunk[:, chunk.shape[1] - args.seq_len: chunk.shape[1]]
targets = chunk[:, chunk.shape[1] - args.seq_len:]
```

open_lm/train.py
```diff
@@ -88,7 +88,7 @@ def sample_chunk(chunk, args):
     if chunk.shape[1] == args.seq_len + 1:
         start_idx = 0
     elif chunk.shape[1] > args.seq_len + 1:
-        start_idx = torch.randint(0, chunk.shape[1] - args.seq_len + 1, (1,)).item()
+        start_idx = torch.randint(0, chunk.shape[1] - args.seq_len - 1, (1,)).item()
```
iejMac, line 91: I think it should be `torch.randint(0, chunk.shape[1] - args.seq_len, (1,)).item()`

jmercat: Oh yeah you're right the high parameter is one above the largest number to be drawn.

iejMac:
LGTM!

cc: @sagadre

jmercat:
Thanks for the review.
btw, the only test I did is train a very small model with sequence length 512 with data tokenized at 2049 sequence length. It failed (at a random iteration) before that change and passed after. Not sure how to properly test this kind of thing.

sagadre:
for testing we can have a small shard with small sequence length and train for enough iterations s.t. we know that all starting indices are hit. but for now lgtm!

sagadre (approved):

PR #109, opened 2023-11-22
support for training AutoModelForCausalLM directly 

sagadre:
support w/ fsdp

tested with mistral 7b:
--hf-model "mistralai/Mistral-7B-v0.1" \
--hf-seq-len 2048 \
--hf-fsdp-block "MistralDecoderLayer" \
--fsdp \
--fsdp-amp

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -44,6 +44,7 @@ except ImportError:
     tensorboard = None
 
 from open_lm.model import create_model
+from open_lm.utils.transformers.hf_wrapper import create_wrapped_hf_model
 from .data import get_data, get_wds_dataset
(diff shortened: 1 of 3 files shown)


open_lm/main.py
```diff
@@ -416,12 +434,29 @@ def main(args):
...
+                        }
+                        break
+
+                if transformer_layer_cls is None:
+                    print(f"--hf-fsdp-block {args.hf_fsdp_block} not found in --hf-model {args.hf_model}")
+                    return -1
+
+            else:
+                transformer_layer_cls = {
+                    Block,
```
achalddave, line 454: nit: can you remove the comma here and on line 444 so Black formats it into one line?

achalddave (approved):

PR #110, opened 2023-11-23
fixed torch_attn that takes dimensions in a different order

jmercat:
[F.scaled_dot_produc_attention documentation](https://pytorch.org/docs/2.0/generated/torch.nn.functional.scaled_dot_product_attention.html?highlight=scaled_dot_product_attention#torch.nn.functional.scaled_dot_product_attention)
[xops.memory_efficient_attention documentation](https://facebookresearch.github.io/xformers/components/ops.html?highlight=memory_efficient_attention#xformers.ops.memory_efficient_attention)

The two different attention functions expect input with different dimension orders. I fixed the torch function input dimension order by calling transpose many times.

The "equivalent" code from the xformers  documentation reads `query @ key.transpose(-2, -1)` ( same as the native pytorch one) but it doesn't match their input dimensions specifications... I [raised an issue for that](https://github.com/facebookresearch/xformers/issues/932#issue-2007439731)

### DIFF
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -88,7 +88,13 @@ def torch_attn(queries, keys, values, is_causal):
     # Need to call contiguous in torch >=2.1, otherwise later calls to .view() fail.
     # Possibly related: https://github.com/pytorch/pytorch/issues/110213 - behavior of scaled_dot_product_attention
     # changed between 2.0 and 2.1
-    return F.scaled_dot_product_attention(queries, keys, values, is_causal=is_causal).contiguous()
+    return (
+        F.scaled_dot_product_attention(
+            queries.transpose(1, 2), keys.transpose(1, 2), values.transpose(1, 2), is_causal=is_causal
+        )
+        .transpose(1, 2)
+        .contiguous()
+    )
 
 
 def get_pos_embed(args: Params):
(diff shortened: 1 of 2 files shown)


iejMac:
great catch! yes this looks about right. Maybe can we get a tiny model training to compare loss curves on a few steps? Afaik this should significantly impact the results for the better when using torch attn.

achalddave:
Oof thanks, good catch. We don't actually use this in training runs, I added it to allow testing on CPU (where xformers doesn't work). 

Wouldn't hurt to have a (gpu) test  comparing different attention variants we use, though. If that's easy for you to add @jmercat that would be great.

iejMac:
just to be clear - the comparison I was talking about was PyTorch attn pre-fix vs. PyTorch attn post-fix. Just to make sure the fix is correct. I think there should be an impact there. But besides that I agree with achal, would be nice to have such a test comparing xformers, PyTorch, or whatever else we decide to support in the future

achalddave:
@jmercat do you think you could add a short test that compares the output of two attentions in this PR? If not, we should still merge this once you resolve conflicts, and open an issue for writing that test in a future PR.

sagadre:
seems like these changes r already in? @achalddave @jmercat  can we close this?

PR #111, opened 2023-11-24
Improved dataloader.

GeorgiosSmyrnis:
Making this PR to get feedback. This changes the dataloader so no tokens are ever repeated for any reason, and also handles failure cases more gracefully.

Still need to implement some tests before merging.


achalddave:
Thanks @GeorgiosSmyrnis! Can you merge or rebase main into your branch so we can see a clean diff for review?

GeorgiosSmyrnis:
My bad - I didn't realize the diffs were not clean.
I'll clean them up and ping again.

GeorgiosSmyrnis:
@achalddave the diffs should be clear now.

open_lm/main.py
```diff
@@ -145,11 +139,13 @@ def load_optimizer(args, model, optimizer, scaler):
...
     checkpoint = pt_load(args.resume, map_location="cpu")
-    if "next_chunk" in checkpoint and "samples_seen" in checkpoint:
-        return checkpoint["next_chunk"], checkpoint["samples_seen"]
+    if "next_shard_per_source" in checkpoint and "samples_seen" in checkpoint:
+        return checkpoint["next_shard_per_source"], checkpoint["samples_seen"]
     else:
-        logging.info(f"=> WARNING: tried to resume a checkpoint without data chunk info. Assuming next_chunk = 0.")
-        return 0, 0
+        logging.info(
+            f"=> WARNING: tried to resume a checkpoint without data chunk info. Assuming next_shard_per_source = 0."
```
achalddave, line 146: Can we leave a more general comment? Like, "Re-starting data loading from the first shard."?

open_lm/file_utils.py
```diff
@@ -175,7 +181,76 @@ def source_exhausted(paths, shard_list_per_source):
...
+            samples_gpu.append(n)
+
+        for worker in range(num_workers):
+            for s in islice(samples_gpu, worker, None, num_workers):
+                samples_per_worker[gpu * num_workers + worker] += s
+
+    return min(samples_per_worker) * num_workers * world_size
+
+
+def count_checkpoints(total_steps, args):
```
achalddave, line 205: This doesn't return anything, should we rename to "log_num_checkpoints()"?

open_lm/main.py
```diff
@@ -707,34 +685,35 @@ def main(args):
...
-            next_chunk=next_chunk if args.dataset_manifest is not None else None,
+            next_shard_per_source=next_shard_per_source if args.dataset_manifest is not None else None,
             samples_seen=samples_seen if args.dataset_manifest is not None else None,
         )
 
-        if final_epoch:
-            logging.info("Ending training due to data exhaustion.")
+        if is_master(args) and done_training:
+            logging.info("Model has seen the desired number of tokens. Ending training.")
             break
```
achalddave, line 717: Should we break in all the workers, not just rank 0?

open_lm/train.py
```diff
@@ -132,14 +134,32 @@ def train_one_epoch(model, data, loss, epoch, step, optimizer, scaler, scheduler
 
     end = time.time()
 
+    try:
+        starting_step = int(optimizer.state_dict()["state"][0]["step"].item())
```
achalddave, line 138: Can we use the step provided to this function instead of relying on optimizer state? (see also https://github.com/mlfoundations/open_lm/pull/102)

open_lm/train.py
```diff
@@ -132,14 +134,32 @@ def train_one_epoch(model, data, loss, epoch, step, optimizer, scaler, scheduler
...
             break
 
+        # If some process no longer has data, stop training - for this epoch
+        procs_with_data = torch.tensor(1, dtype=torch.long, device=device)
+        dist.all_reduce(procs_with_data, op=ReduceOp.SUM)
+        if procs_with_data < args.world_size:
+            other_proc_ran_out_of_data = True
+            break
+
+        if not args.skip_scheduler:
```
achalddave, line 160: I think this is a copy paste error - we call scheduler(step) above on L147.

open_lm/train.py
```diff
@@ -132,14 +134,32 @@ def train_one_epoch(model, data, loss, epoch, step, optimizer, scaler, scheduler
...
         if not args.skip_scheduler:
             scheduler(step)
 
-        if step > total_steps:
+        if step >= total_steps:
             logging.warning(f"step: {step} exceeds total_steps: {total_steps}. ending training.")
             break
 
+        # If some process no longer has data, stop training - for this epoch
+        procs_with_data = torch.tensor(1, dtype=torch.long, device=device)
```
achalddave, line 154: Right now we call dist.all_reduce here and then again at the end of the loop. This is a bit confusing and error-prone. Can we instead do the following?

1. Replace 
    ```
    for i, batch in enumerate(dataloader):
    ```
    with
    ```
    for i in itertools.count():
    ```
    to get an infinite iterator.
2. Add the following here:
    ```
    try:
        batch = next(dataloader)
        has_data = torch.tensor(1, dtype=torch.long, device=device)
    except StopIteration:
        has_data = torch.tensor(0, dtype=torch.long, device=device)
    dist.all_reduce(has_data, op=ReduceOp.SUM)
    if has_data < args.world_size:
        break  # Double check - not sure if we need to do anything else before breaking.
    ```
3. Remove dist.all_reduce call from after the loop.

tests/test_dataset_noResample.py
```diff
@@ -0,0 +1,485 @@
+import pytest
```
achalddave, line 1: can you rename the file to `test_dataset_no_resample.py`?

tests/test_dataset_noResample.py
```diff
@@ -0,0 +1,485 @@
...
+""" TEST STRATEGY:
+Tests to just handle the edge cases in sampling WITHOUT replacement.
+
+We are just going to cover the path where we have:
+- Exactly one source 
+- Never request more data than we have available
+
+And generally we want to assert that:
+- We never see any repeated data 
+- We see exactly the right number of samples as desired (some math here)
```
achalddave, line 41: Can you move this comment to the top of the file?

tests/test_dataset_noResample.py
```diff
@@ -0,0 +1,485 @@
...
+        for line in open(filename, 'r').readlines():
+            total_seq_count += json.loads(line)['num_chunks']
+    return total_seq_count * 2048 if count_tokens else total_seq_count
+
+
+
+def retrieve_dataset_once(total_seqs, paths, epoch, next_shard, weights, seed, disable_buffer, batch_size, num_workers, 
+                          min_shards_needed=1):
+    """ Returns the output of get_wds_dataset -- not dataloader or nothing fancy
+    ONLY WORKS FOR A SINGLE SOURCE
```
achalddave, line 68: nit: no need for all caps since we have an assert below anyway :)

tests/test_dataset_noResample.py
```diff
@@ -0,0 +1,485 @@
...
+    args.dataset_resampled = False
+    args.disable_buffer = disable_buffer
+    args.vocab_size = _MODEL_CONFIGS[args.model]["vocab_size"]
+    args.seq_len = _MODEL_CONFIGS[args.model]["seq_len"]
+    args.world_size = 1 
+    data = get_wds_dataset(args, is_train=True, epoch=epoch, force_num_samples=[total_seqs])
+    return data
+
+
+def retrieve_dataset_once_2(total_seqs, paths, epoch, next_shard, weights, seed, disable_buffer, batch_size, num_workers,
```
achalddave, line 93: maybe rename to a more informative test name?

achalddave:
Thanks for this PR! I left some comments. I haven't been able to get to the tests yet, will try to get to them soon.

achalddave (changes requested):

open_lm/data.py
```diff
@@ -56,6 +57,39 @@ def preprocess_json(text, vocab_size):
...
+        batch.append(sample)
+
+    if len(batch) == 0:
+        return
+    elif len(batch) == batchsize or partial:
+        if collation_fn is not None:
+            batch = collation_fn(batch)
+        yield batch
+    elif len(batch) < batchsize and not partial:
+        for sample in first_batch:
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: This was the intended behavior of batched_fulldata (so that the last few samples are not dropped). Since the number of batches is selected in a way that the last few samples aren't seen, the behavior ends up being essentially the same as wds.batched (and dropping the last batch if partial is False is exactly what wds.batched does, and is its only difference with batched_fulldata).

This function is mostly a leftover from a previous iteration, so we can also just remove it and go back to using wds.batched.

open_lm/data.py
```diff
@@ -56,6 +57,39 @@ def preprocess_json(text, vocab_size):
     return text
 
 
+def _batched_fulldata(data, batchsize=20, collation_fn=wds.filters.default_collation_fn, partial=True):
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: See my response below on this.

GeorgiosSmyrnis:
Thanks @achalddave! Tests are still a bit of WIP, so feel free to disregard them for now. I'll post updates on the PR after some more work is performed on the tests.

open_lm/data.py
```diff
@@ -298,14 +332,20 @@ def filter_lt_seqlen(seq_len, x):
...
 
 
+class FiniteDataPipeline(wds.DataPipeline):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def __iter__(self):
+        """Create an iterator through the pipeline, repeating and slicing as requested.
+
+        This differs from wds.DataPipeline since it allows for slicing even if self.repetitions = 1.
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Yes, `wds.DataPipeline.__iter__` is somewhat counterintuitive. I agree with the docstring and the check - will add them.

open_lm/data.py
```diff
@@ -298,14 +332,20 @@ def filter_lt_seqlen(seq_len, x):
...
+
 def get_wds_dataset(
-    args,
-    is_train,
-    epoch=0,
-    floor=False,
-    tokenizer=None,
-    data_key="json",
-    force_num_samples=None,
+    args, is_train, epoch=0, floor=True, tokenizer=None, data_key="json", force_num_samples=None, multi_epoch=False
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Great point - will add this.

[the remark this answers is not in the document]
GeorgiosSmyrnis: Removed multi_epoch since it doesn't do anything yet.

open_lm/data.py
```diff
@@ -12,12 +12,13 @@
 from multiprocessing import Value
 from functools import partial
 import json
-
 import numpy as np
 import pandas as pd
 import torch
 import webdataset as wds
 from PIL import Image
+from itertools import islice
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: will do

open_lm/data.py
```diff
@@ -354,13 +393,15 @@ def get_wds_dataset(
             pipeline = [wds.SimpleShardList(input_shards)]
 
         # at this point we have an iterator over all the shards
+        # disable shuffling if sampling w/o replacement to ensure no repeat
+        not_use_shuffle = args.disable_buffer or not resampled
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: I agree - will do.

open_lm/data.py
```diff
@@ -435,13 +477,16 @@ def get_wds_dataset(
...
-            num_worker_batches = round_fn(num_batches / num_workers)  # per dataloader worker
+            num_worker_batches = round_fn(all_num_samples[ii] / (global_batch_size * num_workers))
             num_batches = num_worker_batches * num_workers
             num_samples = num_batches * global_batch_size
-            # TODO: what is the effect of setting this?
-            datasets[ii] = datasets[ii].with_epoch(num_worker_batches)  # each worker is iterating over this
+
+            # This forces the dataloader to take num_worker_batches steps per worker, so num_batches total.
+            # Note that this internally sets num_repetitions = sys.maxsize, therefore allowing repeats. We are
+            # safeguarded by the fact that num_worker_batches is the number of minimum worker batches.
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Good idea.

open_lm/file_utils.py
```diff
@@ -6,11 +6,17 @@
 import subprocess
 import sys
 import time
+import logging
+import copy
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Got it, will do.

open_lm/file_utils.py
```diff
@@ -6,11 +6,17 @@
...
+from itertools import islice
 
 import fsspec
 import numpy as np
 import torch
 from tqdm import tqdm
+from itertools import cycle
+
+from .main import is_master
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Got it!

open_lm/file_utils.py
```diff
@@ -175,7 +181,76 @@ def source_exhausted(paths, shard_list_per_source):
     return False
 
 
-def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards_needed):
+def adjust_samples(shard_list, manifest, starting_index, num_workers, world_size):
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Expanded on the docstring of `adjust_samples`.

open_lm/data.py
```diff
@@ -435,13 +477,16 @@ def get_wds_dataset(
         total_num_batches = 0
         total_num_samples = 0
         for ii in range(len(datasets)):
-            num_batches = round_fn(all_num_samples[ii] / global_batch_size)
+            # Calculate batches per worker, round as little as possible.
             num_workers = max(1, args.workers)
-            num_worker_batches = round_fn(num_batches / num_workers)  # per dataloader worker
+            num_worker_batches = round_fn(all_num_samples[ii] / (global_batch_size * num_workers))
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Yeah this makes sense. I'll replace it here, as well as in `file_utils.py`.

open_lm/file_utils.py
```diff
@@ -175,7 +181,76 @@ def source_exhausted(paths, shard_list_per_source):
...
+    """Adjust the number of samples to train on.
+
+    This function returns the number of samples that we will train on, so that no worker will keep going
+    while the others have stopped. This is calculated via the same assignment of nodes that wds.split_by_node
+    and wds.split_by_worker does.
+    """
+    num_samples = [manifest[j]["num_chunks"] for j in range(starting_index, starting_index + len(shard_list))]
+    samples_per_worker = [0 for _ in range(num_workers * world_size)]
+    for gpu in range(world_size):
+        samples_gpu = []
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Yeah it's the same. Replaced it with list slicing.

open_lm/file_utils.py
```diff
@@ -175,7 +181,76 @@ def source_exhausted(paths, shard_list_per_source):
...
+        for n in islice(num_samples, gpu, None, world_size):
+            samples_gpu.append(n)
+
+        for worker in range(num_workers):
+            for s in islice(samples_gpu, worker, None, num_workers):
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Yes, it's the same.

open_lm/file_utils.py
```diff
@@ -175,7 +181,76 @@ def source_exhausted(paths, shard_list_per_source):
     return False
 
 
-def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards_needed):
+def adjust_samples(shard_list, manifest, starting_index, num_workers, world_size):
+    """Adjust the number of samples to train on.
+
+    This function returns the number of samples that we will train on, so that no worker will keep going
+    while the others have stopped. This is calculated via the same assignment of nodes that wds.split_by_node
+    and wds.split_by_worker does.
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Good point. I'll rewrite this function to use `wds.split_by_node` and `wds.split_by_worker directly.

open_lm/file_utils.py
```diff
@@ -175,7 +181,76 @@ def source_exhausted(paths, shard_list_per_source):
...
+                f"checkpointing for the desired amount of epochs, and depends on the number of workers and gpus used, "
+                f"as well as the size of the shards themselves."
+            )
+
+    return
+
+
+def get_string_for_epoch(num_samples, starting_points, paths, weights, num_workers, world_size, multi_epoch=False):
+    if multi_epoch:
+        logging.warning("Multiple passes over the dataset not fully supported yet.")
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Probably the best thing to do. I'll keep the code there for future reference.

open_lm/file_utils.py
```diff
@@ -194,28 +269,84 @@ def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards
...
     return shard_strings_per_source, num_samples_per_source, next_chunk
 
 
-if __name__ == "__main__":
-    print(sys.argv)
-    # path = 's3://s-laion/open_lm_shuffle/rpj_shuffled_mosaic_upsampled_tiktoken_100k_shards_try2/shard_metadata.jsonl'
-    # out = get_string_for_epoch(10000000000 // 2048, 10, path)
+def _single_epoch_string(num_samples, starting_shard_per_source, paths, weights, num_workers, world_size):
+    if len(paths) > 1:
+        logging.warning(
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Agreed.

open_lm/main.py
GeorgiosSmyrnis: Agreed.

open_lm/file_utils.py
GeorgiosSmyrnis: Got it, I'll rename this.

open_lm/main.py
GeorgiosSmyrnis: Good catch!

open_lm/train.py
GeorgiosSmyrnis: My bad, this was an unused artifact from before - I deleted it.

open_lm/train.py
GeorgiosSmyrnis: My bad - removed this.

open_lm/file_utils.py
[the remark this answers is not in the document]
GeorgiosSmyrnis: Actually maybe best to allow this code path for now?
This capability is not fully supported, but some tests where written under the assumption that it exists.

Not sure what's best to do here.

tests/test_dataset_noResample.py
GeorgiosSmyrnis: Changed it to `retrieve_dataset_once_multi_source`.

open_lm/file_utils.py
```diff
@@ -194,28 +269,84 @@ def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards
...
+        except IndexError as e:
+            logging.error(
+                "Number of shards requested for a single epoch is more than the number of shards available. "
+                "Consider lowering the number of workers and / or the number of GPUs."
+            )
+            raise e
+
+    for i in range(len(paths)):
+        # Have the number of shards be a multiple of the total workers.
+        num_multiples = len(shard_list_per_source[i]) // total_num_workers
```
achalddave, line 329: One follow up - I think I don't understand why we need the number of shards to to be a multiple of total workers. Isn't it okay if some workers have more shards than others, since the number of samples per shard may vary anyway?

GeorgiosSmyrnis: This is mostly to avoid throwing away shards - the way webdataset works, each worker goes through a predetermined list of shards (so it cannot say get samples from shards assigned to a different worker or a different node.

We need at least one shard per worker, and it wouldn't cause trouble if we assigned more - but I think this is more convenient (since the leftover shards can be used in the next checkpoint).

open_lm/file_utils.py
```diff
@@ -194,28 +269,84 @@ def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards
...
         shard_strings_per_source.append(shard_string_source)
 
     return shard_strings_per_source, num_samples_per_source, next_chunk
 
 
-if __name__ == "__main__":
-    print(sys.argv)
-    # path = 's3://s-laion/open_lm_shuffle/rpj_shuffled_mosaic_upsampled_tiktoken_100k_shards_try2/shard_metadata.jsonl'
-    # out = get_string_for_epoch(10000000000 // 2048, 10, path)
+def _single_epoch_string(num_samples, starting_shard_per_source, paths, weights, num_workers, world_size):
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Your intuition is correct here - the way this works we can also end up seeing less tokens than requested. I feel this is better than rounding up (worst case you just checkpoint a bit more frequently). I'll update the docstring to reflect this.

For your other point, I'll replace `len(paths)` with `num_sources`.

open_lm/train.py
GeorgiosSmyrnis: I agree, this is far cleaner to what I had before.

However, two calls to dist.all_reduce need to be made (since I'm also communicating the loss value globally, which needs to happen after making sure there is data for it).

open_lm/file_utils.py
achalddave: But with webdataset defaults, some workers can have more shards than others, right? If so, and the number of samples per shard varies, it's not clear to me that ensuring the number of shards is a multiple of number of workers is the right strategy. We could have some imbalanced shard assignments that end up working out in our favor when we adjust_samples, for example [*]. I might be missing something though, please let me know if so!

What we might want instead is to remove shards if one of the workers is going to ignore it completely - but we won't know this until after we call adjust_samples. 

[*] Suppose we have 5 shards with the following number of samples each: [1, 2, 1, 3, 3], and 2 workers. In this case, I believe WDS would assign shards 0, 2, and 4 to worker 1, and shards 1 and 3 to worker 2. If so, ensuring num_shards is a multiple of num_workers would actually lead to suboptimal results: we would save the last shard for the next epoch, and worker 2 would have to drop samples from shard 3 in adjust_samples to match the number of samples in worker 1.

GeorgiosSmyrnis: Yes, webdataset won't complain if you have an imbalanced shard assignment between workers, and I agree that in the above example rounding shards down is not the best thing to do. However, I think the way tokenization works currently most shards have the same size - so this scenario isn't as likely.

I agree that the ideal solution is to verify which shards will be used and which ones won't be. However, it might be cleaner if we just have this requirement (even if suboptimal in some cases).

achalddave: Okay, makes sense. Could you leave a short comment mentioning this motivation? Something like this, or even more concise if you can make it?

```
# Ensure the number of shards is a multiple of number of workers, so each worker has the same
# number of shards.
#
# This is a heuristic to minimize how much data we discard when trying to ensure each worker has
# the same number of samples. Shards tend to have similar number of samples, so an extra shard
# in a worker will likely get discarded.
```

GeorgiosSmyrnis: Sounds good - will do that!

open_lm/file_utils.py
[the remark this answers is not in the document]
GeorgiosSmyrnis: I'll remove `adjust_samples`, since there is no point in keeping only the samples from the smallest worker - it's best to leave them to end their data.

open_lm/file_utils.py
GeorgiosSmyrnis: On a related note, after our discussion offline it seems it's best to not use `adjust_samples` - setting it to the maximum number of samples is pointless. It's best to just let the workers run out of data.

open_lm/file_utils.py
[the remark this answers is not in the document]
GeorgiosSmyrnis: I'll update this to just count the number of samples in these shards, and then let the dataloader workers run out of data. That seems to be the cleanest solution.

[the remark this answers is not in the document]
GeorgiosSmyrnis: Since we no longer need to count the minimum number of shards, I just removed adjust_samples and incorporated its functionality elsewhere.

open_lm/data.py
```diff
@@ -421,6 +452,7 @@ def get_wds_dataset(
         dataset = RandomMix(datasets, probs=args.train_data_mix_weights, longest=True)
```
Vaishaal, line 452: Can we throw a warning here, saying mixing is happening at training time it is preferred to mix during tokenization?

Vaishaal (approved):
mostly looks good, can we improve the error messages to be more meaningful? I left some other comments

open_lm/file_utils.py
Vaishaal: Could we through a warning if the manifest has highly imbalanced shards? The way the tokenize+shuffle code currently works the shard sizes should be approximately equal.

open_lm/data.py
GeorgiosSmyrnis: Included this.

open_lm/file_utils.py
```diff
@@ -200,28 +282,121 @@ def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards
...
+    Args:
+        num_samples: Total number of samples required.
+        starting_shard_per_source: First shard per source that has not been consumed yet.
+        paths: Paths to source manifests.
+        weights: Weighting between sources. If None, it is assumed to be uniform.
+        num_workers_per_gpu: Number of workers per gpu process.
+        world_size: Total number of gpus used for training.
+        factor_drop_lower_avg: While creating the shard list, shards that contain fewer samples than this factor times a
+            running average of samples are skipped. This is done to keep the number of samples per worker uniform, to
+            decrease the amount of elements discarded.
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: This was leftover in the docstring - thanks for catching that!

open_lm/file_utils.py
GeorgiosSmyrnis: I included this check as well. We will throw a warning if a source contains small shards with respect to itself + if sources are imbalanced.

README.md
```diff
@@ -136,6 +136,15 @@ This will create a file called ```manifest.jsonl``` under ```/preproc_data```. T
  --logs path/to/logging/dir/
 ```
 
+### Dataset manifest
+
+The manifest created with `open_lm/utils/make_wds_manifest.py` is a `jsonl` file describing the dataset. Each line in this file corresponds to a shard of the dataset and is a `json` object containing two fields:
+
+- `"shard"`: the name of a shard in the dataset.
+- `"num_chunks"`: the number of sequences contained in the shards. Each sequence contains a set length of tokens.
```
achalddave, line 144: num_sequences?

open_lm/data.py
```diff
@@ -11,13 +11,16 @@
 from dataclasses import dataclass
 from multiprocessing import Value
 from functools import partial
-import json
+from itertools import islice
 
+import json
```
achalddave, line 16: nit: json already imported above

open_lm/data.py
```diff
@@ -36,6 +39,12 @@
 from webdataset.mix import RandomMix
 
 
+def seed_worker(worker_id):
+    worker_seed = torch.initial_seed() % 2**32
```
achalddave, line 43: Do we need to also set the torch seed here? Also, this doesn't use worker_id, why are we taking it as a parameter?

open_lm/file_utils.py
```diff
@@ -181,7 +192,108 @@ def source_exhausted(paths, shard_list_per_source):
     return False
 
 
-def get_string_for_epoch(num_samples, starting_chunk, paths, weights, min_shards_needed):
+def count_small_shards(path, ratio=0.9):
+    shard_sizes = []
```
achalddave, line 196: """Count the number of shards with significantly fewer sequences than the largest shard."""

open_lm/file_utils.py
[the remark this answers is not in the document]
achalddave: okay, fine to leave for now then.

achalddave (approved):

achalddave:
some final nits, and one blocking question re: `worker_seed`. Good to merge after that, thanks for the great PR!

open_lm/data.py
```diff
@@ -451,13 +480,20 @@ def get_wds_dataset(
...
+    else:
+        generator = None
+
     dataloader = wds.WebLoader(
         dataset,
         batch_size=None,
         shuffle=False,
         num_workers=args.workers,
-        persistent_workers=args.dataset_manifest is None,
+        persistent_workers=resampled,
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: What it does is that it keeps the background processes running per checkpoint. This is fine in the resampled case, but it is not something that we best not have in the non replacement case.
Note that resampled and args.dataset_manifest is None are essentially the same.

open_lm/data.py
GeorgiosSmyrnis: Worker_id is taken as a parameter to satisfy the signature of this function, since it is supposed to be called with an `int` as input.

The code was taken from here https://pytorch.org/docs/stable/notes/randomness.html#dataloader: the torch seed is already set beforehand to be correct, this is just to satisfy possible weirdness when using other libs.

GeorgiosSmyrnis:
Thanks a lot @achalddave ! I made the minor edits and looked into worker_init_fn. I'll merge after CI passes.

PR #112, opened 2023-11-24
stats file and better handling of checkpoint saving

sagadre:


GeorgiosSmyrnis (approved):
LGTM!

open_lm/main.py
```diff
@@ -212,6 +228,9 @@ def save_checkpoint(
             previous_checkpoint = os.path.join(args.checkpoint_path, f"optimizer_{completed_epoch - 1}.pt")
             if os.path.exists(previous_checkpoint):
                 os.remove(previous_checkpoint)
+            previous_checkpoint = os.path.join(args.checkpoint_path, f"stats_{completed_epoch - 1}.pt")
```
achalddave, line 231: Nit: can we make this a loop over epoch, optimizer and stats to avoid potential copy paste errors now and in the future?

open_lm/main.py
```diff
@@ -729,12 +750,13 @@ def main(args):
             completed_epoch,
             evaluation_loss,
             step=global_step,
+            is_final_checkpoint=final_epoch or (completed_epoch == args.epochs),
```
achalddave, line 753: Can be a follow up PR, but should we just set args.epoch to completed_epoch earlier (and emit a warning) instead of creating an auxiliary final_epoch variable?

open_lm/main.py
```diff
@@ -195,7 +196,18 @@ def save_checkpoint(
         if scaler is not None:
             checkpoint_dict_opt["scaler"] = scaler.state_dict()
 
-        if completed_epoch == args.epochs or (args.save_frequency > 0 and (completed_epoch % args.save_frequency) == 0):
+        checkpoint_dict_stats = {
+            "epoch": completed_epoch,
+            "name": args.name,
+            "is_final_checkpoint": is_final_checkpoint,
+            "evaluation_loss": evaluation_loss,
+        }
```
achalddave, line 204: We're not using these stats right now, right? We just create them for inspection?

achalddave:
Left minor comments. Feel free to merge if blocking and address in a follow up PR

open_lm/main.py
sagadre: setting args.epoch a bit dicey as we are using this var in an outer for loop. but agree this extra var prob is not the best either. we can do something different in a later PR

PR #118, opened 2023-11-28
Added some tests for LR scheduling

revbucket:
Just a few tests added to `test_training_sample.py` which check that we're modifying LR correctly. 

Also added some optional params (`finegrain-debug`) that, if passed, cause `main.main(...)` to exit directly before we start looping over epochs, returning all local vars (for finegrained debugging purposes, obv)


achalddave:
Sorry for the super late review here. One main question: Is there a way we can run all of main (including its call to train_one_epoch), and only inspect the LR after that? This way, we can catch potential errors in how main calls train_one_epoch. What do you think?

revbucket:
Okay, I added better tests for `main.py`, just directly asserting on the tensorboard logger's outputs

achalddave:
This looks great, thanks! Can you black format, merge in changes with main, and ensure the CI runs and passes?

tests/test_training_simple.py
```diff
@@ -14,3 +42,265 @@ def test_train_simple():
...
+            assert all(abs(lr - compute_lr(global_step - 1)) < 1e-8 for lr in lrs)
+
+
+
+# =========================================================
+# =                main.py tests for LR                   =
+# =========================================================
+
+
+def test_lr_scheduling_from_main():
```
achalddave, line 265: Can we just use this style of tests for all tests (read from TB without finegrain_debug)? It'll simplify the params and main and we don't need finegrain_debug.

achalddave (approved):
This is awesome, thanks @revbucket! Once you fix the conflicts with main and make sure CI passes, this is good to merge. Leaving an approval to avoid blocking on me once this is done.

PR #119, opened 2023-11-29
Allowing datapreprocess to be accessible as (sub)package

jeffreywpli:
Previously only stuff in `open_lm` dir is included when building a package with `setup.py`. This PR adds `__init__.py`'s  to `datapreprocess` and moves it within `open_lm` to allowing access to things like tokenize_shuffle as a subpackage of `open_lm`


Vaishaal (approved):
LGTM

Issue #121, opened 2023-11-30
Move dummy cred download into test


achalddave:
Tests currently fail locally because we download credentials in the workflow rather than in the tests. We should move this into the tests:

https://github.com/mlfoundations/open_lm/blob/b27ce1e26db571f3a6558e126a0d1cc98b58908d/.github/workflows/ci.yml#L46C1-L46C173

PR #122, opened 2023-11-30
Reincorporate manifest creation.

GeorgiosSmyrnis:
Add manifest creation in the tokenization script with the new format. The manifests made by this will be compatible with the new dataloader version.

### DIFF
--- a/open_lm/datapreprocess/ray/tokenize_shuffle.py
+++ b/open_lm/datapreprocess/ray/tokenize_shuffle.py
@@ -277,11 +277,8 @@ def map_write_wds(batch, batch_size, folder, counter):
     bio.seek(0)
     token_count = ray.get(counter.increment_token_count.remote(token_count))
     write_to_location(folder, tar_name, bio)
-    
-    return_dict = {
-        "shard": [tar_name.split(".")[0]],
-        "num_sequences": [len(batch["tokens"])]
-    }
+
+    return_dict = {"shard": [tar_name.split(".")[0]], "num_sequences": [len(batch["tokens"])]}
 
     return return_dict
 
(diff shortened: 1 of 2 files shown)


GeorgiosSmyrnis:
This now also addresses #121 for the tests.

open_lm/datapreprocess/ray/tokenize_shuffle.py
```diff
@@ -367,6 +372,27 @@ def glob_files(path, suffix=".jsonl"):
     return matching_files
 
 
+def get_filesystem(environment):
```
Vaishaal, line 375: This won't work in most cases

open_lm/datapreprocess/ray/tokenize_shuffle.py
```diff
@@ -484,7 +510,24 @@ def main(args):
             "counter": counter,
         },
         batch_format="pandas",
-    ).count()
+    )
```
Vaishaal, line 513: Instead of a count, do a .take_all()  which should return O(shards) files which is the size of the manifest and then just write directly to s3 using boto3. It will be cleaner, you can even write the manifest to disk
```
import boto3
import json

def upload_jsonl_to_s3(bucket_name, s3_key, data):

    s3_client = boto3.client('s3')


    jsonl_content = '\n'.join(json.dumps(record) for record in data)


    response = s3_client.put_object(
        Bucket=bucket_name,
        Key=s3_key,
        Body=jsonl_content
    )


    if response['ResponseMetadata']['HTTPStatusCode'] == 200:
        print("Upload successful.")
    else:
        print("Upload failed.")
data = [{'key1': 'value1'}, {'key2': 'value2'}]
upload_jsonl_to_s3('my-bucket', 'path/to/myfile.jsonl', data)


```

Vaishaal (changes requested):

open_lm/datapreprocess/ray/tokenize_shuffle.py
GeorgiosSmyrnis: Removed this function.

open_lm/datapreprocess/ray/tokenize_shuffle.py
GeorgiosSmyrnis: Thanks for the suggestion! Changed this.

Vaishaal (approved):

PR #125, opened 2023-11-30
Fixes for gradient accumulation test

achalddave:
Earlier, we used deepcopy, which seems to copy the tensor python objects, but keeps a pointer to the same tensor, meaning we were not properly testing gradient accumulation. This commit mainly fixes that, using load_state_dict() instead to make a model copy.

Unfortunately, the test fails once we do this, and a few changes are necessary to make it pass:

1. Switch AdamW -> SGD
2. Switch to float32

It's unclear why this was necessary, and solving this is left for a future commit.


achalddave:
Created https://github.com/mlfoundations/open_lm/issues/126 to keep track of the AdamW + grad accum issue.

GeorgiosSmyrnis (approved):
LGTM - let's keep track of the issue #126 and see what's going on.

PR #127, opened 2023-12-01
Support for initializing models on meta device

achalddave:
Allows loading much larger models without using CPU ram, only initializing after we FSDP and shard onto GPU. 

Earlier, we were limited in model size by the memory of a single GPU, due to this line:
https://github.com/mlfoundations/open_lm/blob/005fdccca337f5a84c99bbfecee768a77e90bd9e/open_lm/main.py#L396

Removing that line limits us by the CPU memory of a single node, even in multi-node settings, but this is still not enough for large models (e.g., 34B takes >1TB RAM, crashes a p4).

Now, we are limited by the total GPU memory across nodes, so we can load large models in multi-node setting.
Our model size is now limited by total GPU memory across nodes.

Note: This first requires merging #125

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -394,6 +394,8 @@ def main(args):
     if args.val_num_samples is not None:
         args.val_num_samples //= args.seq_len
 
+    random_seed(args.seed, args.rank)
+
     if args.distributed:
         if args.fsdp:
             transformer_layer_cls = None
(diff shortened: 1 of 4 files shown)


GeorgiosSmyrnis (approved):
Looks good overall, but if you can rebase to the newest main branch before merging that would be great (just to make this PR clearer)!

achalddave:
Note: One remaining issue is we load and save checkpoints from a single machine, when we should be using https://pytorch.org/docs/stable/distributed.checkpoint.html. This will be addressed in a future PR.

PR #129, opened 2023-12-03
tokenize_shuffle.py: tar files as input format

iejMac:


iejMac:
replacing https://github.com/mlfoundations/open_lm/pull/66/files

open_lm/datapreprocess/ray/tokenize_shuffle.py
```diff
@@ -313,6 +341,9 @@ def load_tokenizer(tokenizer):
     if tokenizer == "EleutherAI/gpt-neox-20b":
         enc = GPTNeoXTokenizerFast.from_pretrained("EleutherAI/gpt-neox-20b")
         return (lambda x: enc(x).input_ids, enc.vocab_size)
+    elif tokenizer == "none":
+        enc = lambda x: x
+        return (enc, -1)
```
iejMac, line 346: not sure what to do for this. Maybe we can have the "tokenizer" arg double as a "vocab_size" arg for pre-tokenized data and then we just check `elif tokenizer.isnum()`. Not sure if this is clean

iejMac:
this is pretty much good to go minus that one point about the "vocab_size" output of load_tokenizer

cc: @GeorgiosSmyrnis @Vaishaal

open_lm/datapreprocess/ray/tokenize_shuffle.py
iejMac: I tried something else, lmk how that it works:

If vocab_size is ever passed in that must mean tokenizer won't be needed so we don't load the tokenizer and instead return (lambda x: x, args.vocab_size)

Vaishaal:
lgtm!

Vaishaal (approved):
LGTM

PR #137, opened 2023-12-06
model.py: Mamba SSM

iejMac:
paper: https://arxiv.org/pdf/2312.00752.pdf
comparison: https://api.wandb.ai/links/iejmac/xa3q94vr

Command for the run:
```
torchrun --nproc-per-node 8 -m open_lm.main \
        --train-data "pipe:aws s3 cp s3://laion-west/rpj_tokenized_upsampled_eleutherai/shard_{00000000..00099999}.tar -" \
        --train-num-samples 1600000000 \
        --dataset-resampled \
        --workers 1 \
        --batch-size 8 \
        --accum-freq 1 \
        --log-every-n-steps 1 \
        --grad-clip-norm 1 \
        --lr 3e-3 \
        --lr-cooldown-end 3e-4 \
        --wd 0.1 \
        --model "open_lm_160m" \
        --precision amp_bfloat16 \
        --beta2 0.95 \
        --warmup 50 \
        --epochs 1 \
        --delete-previous-checkpoint \
        --data-key "json" \
        --wandb-project-name open-lm \
        --report-to wandb \
```


open_lm/model.py
```diff
@@ -305,20 +309,42 @@ def create_params(args):
...
+            norm_type=get_norm_class(cfg.get("model_norm", args.model_norm)),
+            apply_qk_norm=cfg.get("qk_norm", args.qk_norm),
+            positional_embedding_type=cfg.get("positional_embedding_type", args.positional_embedding_type),
+            ffn_type=cfg.get("ffn_type", args.ffn_type),
+        )
 
 
 def create_model(args):
-    return Transformer(create_params(args))
+    if 'mamba' in args.model:
```
achalddave, line 337: This is fine for now, but would be nice if we could minimize how many places we have if/else. Maybe create_params(args) and then have a Mamba subclass that takes that in and handles all the modifications (vocab_size, popping seqlen, etc.)

achalddave:
This is _excellent_. These plots from your wandb are super impressive. 
image
image

Left a couple small comments, if you can reformat I think we can merge.

open_lm/model.py
```diff
@@ -305,20 +310,52 @@ def create_params(args):
...
+
+class OpenLMMamba(nn.Module):
+    # Experimental architecture, please "pip install mamba-ssm"
+    # https://arxiv.org/abs/2312.00752
+    def __init__(self, params):
+        super().__init__()
+        self.seq_len = params.pop("seq_len")
+        self.vocab_size = params["vocab_size"]
+
+        self.model = MambaLMHeadModel(**params)
```
sagadre, line 345: consider checking if MambaLMHeadModel is None here and printing/logging the instructions needed to install this package. Also do we wanna do a readme update with the relevant installs?

open_lm/model.py
```diff
@@ -305,20 +310,52 @@ def create_params(args):
...
+            weight_tying=cfg["weight_tying"],
+            norm_type=get_norm_class(cfg.get("model_norm", args.model_norm)),
+            apply_qk_norm=cfg.get("qk_norm", args.qk_norm),
+            positional_embedding_type=cfg.get("positional_embedding_type", args.positional_embedding_type),
+            ffn_type=cfg.get("ffn_type", args.ffn_type),
+        )
+
+
+class OpenLMMamba(nn.Module):
+    # Experimental architecture, please "pip install mamba-ssm"
```
sagadre, line 338: nit: we can call this Mamba to be consistent with the Transformer class we use for other open_lm models

open_lm/model.py
iejMac: done! I think readme update not required given this architecture is still experimental

sagadre:
thanks lgtm!

sagadre (approved):

sagadre:
in a follow-up PR we can look more at fsdp + mamba

PR #138, opened 2023-12-06
fix load_model module key edit

afang-story:
https://github.com/mlfoundations/open_lm/pull/127
Above PR changed order from model -> load -> distributed to model -> distributed -> load

load_model automatically removes "module" from keys so above change can break load_model
also breaks if distributed and loading weights that do not include module in keys
this pr check if distributed to decide whether to load state_dict to model or model.module


Vaishaal (approved):
LGTM

PR #141, opened 2023-12-07
Allow parsing parameters from a config file

achalddave:
This allows us to load parameters from a json or yaml file. If a config is passed via file, then no other arguments are allowed, to avoid confusion due to commandline overrides. Added a few tests that check for equality between config and command line args.

### DIFF
--- a/open_lm/params.py
+++ b/open_lm/params.py
@@ -2,6 +2,7 @@ import argparse
 import ast
 import copy
 import json
+import logging
 import yaml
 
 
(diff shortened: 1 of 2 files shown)


open_lm/params.py
```diff
@@ -61,6 +64,59 @@ def add_model_args(parser):
...
+
+def maybe_load_config(parser, args):
+    config_parser = argparse.ArgumentParser()
+    config_parser.add_argument("--config", type=str)
+    args, unknown_args = config_parser.parse_known_args(args)
+    if not args.config:
+        return None
+
+    assert not unknown_args, "No arguments can be passed if --config is provided."
+    print(f"Loading config from: {args.config}")
```
sagadre, line 93: nit: print or use logger? maybe does not matter here

open_lm/params.py
```diff
@@ -61,6 +64,59 @@ def add_model_args(parser):
...
+    with open(args.config, "r") as f:
+        if args.config.endswith(".yaml"):
+            config = yaml.safe_load(f)
+        else:
```
sagadre, line 97: maybe explicitly check for json then raise if not yml or json?

achalddave:
thanks! addressed, merging now

PR #147, opened 2023-12-11
Vaishaal/revert fsdp loading

Vaishaal:


Vaishaal:
Commits make resuming from checkpoint broken. Needed for larger (34B) models, but not now.

achalddave (approved):

Issue #145, opened 2023-12-11
Add test for checkpoint loading after save


achalddave:
We should add a test that:
1. trains a (small) model for a couple steps
2. saves it to disk
3. calls main() again with a path to the checkpoint on disk
4. trains a few steps

The test should test single process, DDP, and FSDP.

jmercat:
I cannot assign this to myself but I can do it

achalddave:
Closing this issue as we have added a test in #148, but @Vaishaal feel free to open a new issue if you reproduce the checkpoint loading bug you have

PR #148, opened 2023-12-11
Test train save load, with and without fsdp

jmercat:
Should help address #145, please tell me if I should actually call main as suggested in the issue or if this is fine.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -217,9 +217,11 @@ def save_checkpoint(
             or (args.save_frequency > 0 and (completed_epoch % args.save_frequency) == 0)
         ):
             for prefix in prefixes:
+                path = os.path.join(args.checkpoint_path, f"{prefix}{completed_epoch}.pt")
+                print(f"Saving {prefix}{completed_epoch} in {path}...")
                 torch.save(
                     prefixes[prefix],
-                    os.path.join(args.checkpoint_path, f"{prefix}{completed_epoch}.pt"),
+                    path,
                 )
 
         if args.delete_previous_checkpoint:
--- a/tests/shared.py
+++ b/tests/shared.py
@@ -2,26 +2,26 @@ import torch
 from torch import optim
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 
+from open_lm.data import get_data
+from open_lm.distributed import init_distributed_device
 from open_lm.main import random_seed
 from open_lm.model import create_model
-from open_lm.data import get_data
 from open_lm.scheduler import cosine_lr
 from tests.utils import download_val_data
 
 
 class MockTrainArgs:
-    def __init__(self, model):
+    def __init__(self, model, **kwargs):
         data_path = download_val_data("shard_00000000.tar", "./tests/assets/")
 
         self.model = model  # part of model config
         self.model_norm = "gain_only_layer_norm"
-        self.rotary_old = False
         self.qk_norm = False
         self.train_data = [
             data_path,
         ]
         self.log_logit_mean = False
-        self.device = 0
+        self.device = "cpu"
         self.precision = "float32"
         self.wd = 0.033
         self.lr = 3e-3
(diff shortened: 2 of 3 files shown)


open_lm/main.py
```diff
@@ -226,9 +226,11 @@ def save_checkpoint(
             or (args.save_frequency > 0 and (completed_epoch % args.save_frequency) == 0)
         ):
             for prefix in prefixes:
+                path = os.path.join(args.checkpoint_path, f"{prefix}{completed_epoch}.pt")
+                print(f"Saving {prefix} {completed_epoch} in {path}...")
```
achalddave, line 230: `{prefix}{completed_epoch}` (without space) in print?

tests/shared.py
```diff
@@ -36,6 +37,9 @@ def __init__(self, model):
         self.rank = 0
         self.local_rank = 0
         self.log_every_n_steps = 1e8
+        self.save_logs = True
```
achalddave, line 40: maybe keep `save_logs False` by default and we can set it in just the tests where we care about logs?

tests/shared.py
```diff
@@ -106,9 +116,10 @@ def create_train_fixtures(model="open_lm_11m", fsdp=False):
     # create base models
     random_seed()
     if fsdp:
+        init_distributed_device(args)
         with torch.device("meta"):
             model = create_model(args)
-        model = FSDP(model)
+            model = FSDP(model)
```
achalddave, line 122: Hm, was this necessary? I think actually the FSDP() call should be outside the meta device, so that FSDP materializes the tensors. 

Though really, since we've reverted the meta tensor commit, we should really just remove the `with torch.device("meta")` call here. I'll add an issue for this for a separate PR

tests/test_save_load.py
```diff
@@ -0,0 +1,114 @@
...
+    args2, model2, data2, optimizer2, scheduler2, loss2 = create_train_fixtures(tiny_args.model, fsdp=False)
+    args.vocab_size = tiny_open_lm.config.vocab_size
+    args.seq_len = tiny_open_lm.config.seq_len
+    args.train_num_samples = 16
+    args.batch_size = 4
+    args.checkpoint_path = "./tests/assets/checkpoints/tiny_model/"
+    os.makedirs(args.checkpoint_path, exist_ok=True)
+
+    print("Training tiny model")
+    train_one_epoch(tiny_open_lm, data, CrossEntropyLossWithZLoss(), epoch=epoch, step=global_step, optimizer=optimizer, scaler=scaler, scheduler=scheduler, total_steps=-1, args=args)
```
achalddave, line 60: black format?

tests/test_save_load.py
```diff
@@ -0,0 +1,114 @@
...
+def test_tiny_save_load_fsdp(tiny_open_lm, tiny_args):
+    world_size = 1
+    mp.spawn(
+        _save_load_helper_fsdp,
+        args=(world_size, tiny_open_lm, tiny_args),
+        nprocs=world_size,
+        join=True
+    )
+
+# def test_s3_load(
```
achalddave, line 111: remove?

achalddave (commented):
Looks great! couple nits, left some small comments.

tests/shared.py
achalddave: Actually, can you go ahead and remove the `with torch.device("meta")` call while you're at it? It shouldn't break anything. If it makes some tests fail, feel free to leave it in for now and we can revisit later.

tests/shared.py
jmercat: If set to false my understanding is that it doesn't save the checkpoint at all. BTW is that a bug?

achalddave: i see okay, that is weird. https://github.com/mlfoundations/open_lm/blob/73e7b3709057d62408d700dcaad4ebc6317c56b9/open_lm/main.py#L610 should set save_logs to True if you've specified a logs dir..

jmercat: I set it to false in the shared file but I did have to set it to true in my test otherwise the fsdp test would not pass (not sure why).

achalddave (approved):

Issue #55, opened 2023-10-24
Tokenization on-the-fly without slowdown


sagadre:
Benchmark and get tokenization on-the-fly to be as fast as training on pre-tokenized data

PR #150, opened 2023-12-11
Replace batch_size with global_batch_size.

GeorgiosSmyrnis:
This PR deprecates the per gpu batch size from the command line and replaces it with global batch size.

Tests are also updated to account for this.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -247,7 +247,7 @@ def main(args):
         args.global_batch_size % args.world_size == 0
     ), f"Global batch size ({args.global_batch_size}) is not divisible by number of GPUs ({args.world_size}), and thus cannot be respected."
 
-    args.per_gpu_batch_size = args.global_batch_size // args.world_size
+    args.per_gpu_batch_size = max(args.global_batch_size // args.world_size, 1)
     if args.val_data is not None:
         args.per_gpu_val_batch_size = max(args.global_val_batch_size // args.world_size, 1)
 
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -249,7 +249,7 @@ def main(args):
 
     args.per_gpu_batch_size = args.global_batch_size // args.world_size
     if args.val_data is not None:
-        args.per_gpu_val_batch_size = args.global_val_batch_size // args.world_size
+        args.per_gpu_val_batch_size = max(args.global_val_batch_size // args.world_size, 1)
 
     if args.hf_model is not None and args.hf_seq_len is None:
         raise ValueError("If passing --hf-model, must also pass --hf-seq-len to be used for training/fine-tuning.")
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -248,6 +248,8 @@ def main(args):
     ), "Global batch size is not divisible by number of GPUs, and thus cannot be respected."
 
     args.per_gpu_batch_size = args.global_batch_size // args.world_size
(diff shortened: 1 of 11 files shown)


open_lm/params.py
```diff
@@ -576,7 +576,7 @@ def parse_args(args):
 
     if args.val_data is not None and args.val_batch_size is None:
         # if not set explicitly make sure that the val batch size is set to the micro batch size
-
-        args.val_batch_size = args.batch_size // args.accum_freq
+        # TODO: is this correct with global batch size?
```
GeorgiosSmyrnis, line 579: This part I'm not 100% sure about, I think it is assumed that we are running eval with only 1 GPU.

GeorgiosSmyrnis:
This currently fails on the GPU tests, so it needs some changes.

open_lm/params.py
achalddave: I would do

```args.per_gpu_batch_size_val = (args.global_batch_size // world_size // args.accum_freq)```

and change references to `args.val_batch_size` to `args.per_gpu_batch_size_val` (I know, it's a mouthful). 

Other than that this LGTM!

**Comment by jmercat** [comment]
I think we might want to consider having a global batch size in terms of tokens instead of in terms of samples. In most work I saw this is the thing that is kept constant regardless of the sequence length… might be unintuitive to the user though… what do you think?

File: open_lm/main.py [hunk 2]
```diff
@@ -243,6 +243,15 @@ def main(args):
...
     device = init_distributed_device(args)
 
+    assert (
+        args.global_batch_size % args.world_size == 0
+    ), "Global batch size is not divisible by number of GPUs, and thus cannot be respected."
+
+    args.per_gpu_batch_size = args.global_batch_size // args.world_size
+    if args.val_data is not None:
+        # Make sure that val batch size is set to micro batch size
+        args.per_gpu_val_batch_size = args.global_val_batch_size // args.world_size // args.accum_freq
```
rom1504, line 253: Should accum freq be checked ?

GeorgiosSmyrnis:
@jmercat I think the batch size in terms of sequences is a better choice, since we assume the data to already be tokenized / split into sequences - if we chose the batch size to be in terms of tokens, the only valid choices are multiples of the chosen sequence length, which I think would make things fairly unintuitive.

Maybe we would need to rethink this in the future, but in order to fully support this we would need to essentially resolve #55 first I think.

open_lm/main.py
```diff
@@ -243,6 +243,15 @@ def main(args):
     # fully initialize distributed device environment
     device = init_distributed_device(args)
 
+    assert (
```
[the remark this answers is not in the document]
GeorgiosSmyrnis: Good point! Added!

open_lm/main.py
GeorgiosSmyrnis: I changed how `accum_freq` is handled to make it more similar to how it was handled before (in `params.py`). I think this skips the need for checks, unless you have something specific in mind.

achalddave: This looks fine, but just realized, maybe just add a max(..., 0) to ensure we don't accidentally get batch size of 0?

achalddave (approved):
Left one comment + need to sync with main, feel free to merge after that!

sagadre:
can we get a version bump on this?

sedrick-keh-tri:
Old thread, but just curious. What's the motivation behind switching from per gpu batch size to global batch size?

PR #151, opened 2023-12-12
[DRAFT] Sagemaker integration

sedrick-keh-tri:
This still needs some fixing because there are some TRI-specific stuff in launch_sagemaker_train.py 

It works if you run `python sagemaker_train/launch_sagemaker_train.py --user sedrick.keh --cfg-path sagemaker_train/cfg_sample.yaml --build`. For easy testing, I edited params.py to accept openlm_mix_tri_s3 in my local code (not pushed) but aside from that it works without any other changes


sedrick-keh-tri:
Edited the file to remove TRI mentions. Now the region, ARN, and s3 path can be supplied through command line or env variable

sagemaker_train/cfg_sample.yaml
```diff
@@ -0,0 +1,37 @@
...
+# grad-checkpointing: False
+grad-clip-norm: 1
+log-every-n-steps: 20
+model: "open_lm_7b"
+name: "sample_7b"
+precision: "amp_bfloat16"
+report-to: "wandb"
+seed: 124
+train-data-mix-weights: 0.725 0.275
+train-data: "openlm_mix_tri_s3"
```
achalddave, line 20: let's change this to "TODO". Also should it be a list?

sagemaker_train/launch_sagemaker_train.py
```diff
@@ -0,0 +1,198 @@
...
+
+    ##########
+    # Configure the training
+    ##########
+    base_job_name = f"{args.user.replace('.', '-')}-{NAME}"
+
+    checkpoint_local_path = "/opt/ml/checkpoints"
+
+    with open(args.cfg_path, "r") as f:
+        train_args = yaml.safe_load(f)
```
achalddave, line 137: Now that openlm supports a --config, let's just pass the config to openlm directly, with --config args.cfg_path?

achalddave (commented):
small nits, otherwise looks good to me!

sagemaker_train/launch_sagemaker_train.py
sedrick-keh-tri: Tried this but it seems that it gives some typing errors when passed through sagemaker:
`Type mismatch (config: <class 'str'> vs. argparse: <class 'bool'>) with values (config:  vs. argparse: False) for config. key: dataset_resampled`

Leaving it as is for now

sagemaker_train/cfg_sample.yaml
```diff
@@ -0,0 +1,37 @@
+accum-freq: 4
+beta1: 0.9
+beta2: 0.95
+data-key: "json"
+dataset-resampled: ""
```
achalddave, line 5: If you set this to True instead of empty string, the error you mentioned should go away when passing the config via path (and similarly for all other keys which have a "" value - set them to True instead)

sedrick-keh-tri: I tried that and it still gives the same error. It seems to read everything as a string.

Just to double-check: The way to pass a config is to just do `train_args = {"config": args.cfg_path}` instead of the yaml.safe_load(f), right?

achalddave: Hmm yeah that should be all you need. And you rebuilt the docker container after that change right? Will try it out later today, maybe there's something wrong with the parsing logic.

achalddave:
made some changes that should fix the issues with config here: https://github.com/sedrick-keh-tri/open_lm/pull/1. I think if you merge that PR into your branch, it should update this PR, and we should be good to merge.

achalddave (approved):

PR #149, opened 2023-12-11
global batch size and val fix

sagadre:

PR #152, opened 2023-12-12
Val list -> tensor bugfix

achalddave:
This is just the bugfix part of Samir's PR here https://github.com/mlfoundations/open_lm/pull/149, without the global batch size part.


sagadre (approved):
thanks!

PR #162, opened 2023-12-18
Fix some extra cases in global batch size.

GeorgiosSmyrnis:
This fixes some issues caused by #150 - one that broke `log_num_checkpoints.py`, as well as some related to test cases.


sagadre (approved):
lgtm! thanks!

PR #164, opened 2023-12-18
Call reset_params explicitly in rotary embedding

achalddave:
This is necessary for supporting `--disable-meta-device` properly. This shouldn't have affected our runs, since we used meta device by default, but improves our support for disabling meta device.


achalddave:
Closed in favor of #165

PR #165, opened 2023-12-18
Disable meta device by default

achalddave:
The meta device initialization seems to have some issues - for example, FSDP does not seem to call reset_parameters on all our modules. Given this, and the fact that we currently aren't targeting models above 34B, let's disable the meta device entirely for now.


GeorgiosSmyrnis (approved):
LGTM, thanks!

PR #167, opened 2023-12-19
add non linear attention following https://arxiv.org/abs/2309.08586

sagadre:


tests/test_relu_attention.py
```diff
@@ -0,0 +1,58 @@
...
+        [1, 1024, 2048, 8, 128],
+    ]:
+        queries = torch.rand(bs, q_seq_len, h, d)
+        keys = torch.rand(bs, k_seq_len, h, d)
+        values = torch.rand(bs, k_seq_len, h, d)
+
+        for is_causal in [True, False]:
+            torch_out = torch_attn(queries, keys, values, is_causal=is_causal)
+
+            my_out = non_linear_attn(
```
mitchellnw, line 20: I like this test! What about also testing equivalence to xformers attn?

open_lm/attention.py
```diff
@@ -0,0 +1,143 @@
...
+            )
+            .transpose(1, 2)
+            .contiguous()
+        )
+
+
+ATTN_NON_LINEARITIES = {
+    "relu": F.relu,
+    "relu_squared": lambda x: torch.pow(F.relu(x), 2),
+    # "gelu": F.gelu, # goes to NaN with bais so comment out for now
```
mitchellnw, line 76: Not a big deal, but could explicitly just 0 out the illegal attn weights instead of adding -inf.

mitchellnw (commented):
Looks great! Once we see training curves we can probably feel confident to approve/merge.

sagadre:
curves matching for cases of xformers softmax attention and softmax attention going throught the custom attention code path
Screen Shot 2023-12-20 at 12 23 21 AM

sagadre:
relu^2 with alpha = 1 and sequence length scalar determined by max sequence length or avg sequence length approaching the standard softmax setting. for context, this is an 11M param model trained for 220M tokens. larger runs planned for future dates, but probably not blocking this PR

Screen Shot 2023-12-20 at 11 02 51 AM

mitchellnw (approved):
LGTM!

PR #175, opened 2023-12-20
moe requirements + readme

sedrick-keh-tri:
- Updated some params to match the current param format in the main branch
- Added megablocks to requirements.txt (Note: There seems to be an issue with the PyPI package so I'm building from source here. See https://github.com/stanford-futuredata/megablocks/issues/67)


sedrick-keh-tri:
Hmm, ok I can't seem to get this to work. Maybe let's just hold off on this until the PyPI package gets fixed

PR #153, opened 2023-12-13
add parameter compatability check

IgorVasiljevic-TRI:
https://github.com/mlfoundations/open_lm/issues/107

PR #176, opened 2023-12-20
Fix checkpoint loading and checkpoint tests

### LINKED ISSUE #170: Use distributed when world_size=1 if requested
Right now, we disable distributed functionality if we are using a single gpu: https://github.com/mlfoundations/open_lm/blob/main/open_lm/distributed.py#L20-L25. But if WORLD_SIZE is provided, we should behave as if we are in a distributed environment, which in turn will allow us to run tests that verify the distributed code paths without requiring multiple gpus.

achalddave:
This addresses a few issues with our checkpoint save/load tests, and with checkpoint loading itself:

1. Adds a new test that calls main() to test checkpoint saving and loading, ensuring that our full main() pipeline works correctly.
2. Adds a `--force-distributed` flag that allows us to test distributed code paths on a single gpu. Previously, all our tests were running with args.distributed=False, even when args.fsdp=True.
3. Reverts a bug introduced by #153, which meant we did not automatically resume the latest checkpoint when using s3.
4. Fixes a bug in `--resume latest` which led to checkpoints being ignored when resuming in the case of local checkpoints without remote-sync.

Closes #170

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -109,7 +109,9 @@ def load_model(args, model):
         global_step = checkpoint.get("step", None)
         if next(iter(sd.items()))[0].startswith("module"):
             sd = {k[len("module.") :]: v for k, v in sd.items()}
-        if args.distributed:
+        if args.fsdp:
+            model.load_state_dict(sd)
+        elif args.distributed:
             model.module.load_state_dict(sd)
         else:
             model.load_state_dict(sd)
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -81,7 +81,7 @@ def natural_key(string_):
 def get_latest_checkpoint(path: str):
     is_s3 = path.startswith("s3")
     fs, root_path = fsspec.core.url_to_fs(path)
-    checkpoints = fs.glob(os.path.join(root_path, "**/epoch_*.pt"))
+    checkpoints = fs.glob(os.path.join(root_path, "epoch_*.pt"))
     if checkpoints:
         checkpoints = sorted(checkpoints, key=natural_key)
         return f"s3://{checkpoints[-1]}" if is_s3 else checkpoints[-1]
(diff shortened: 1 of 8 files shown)


open_lm/main.py
```diff
@@ -81,7 +81,7 @@ def natural_key(string_):
 def get_latest_checkpoint(path: str):
     is_s3 = path.startswith("s3")
     fs, root_path = fsspec.core.url_to_fs(path)
-    checkpoints = fs.glob(os.path.join(root_path, "**/epoch_*.pt"))
+    checkpoints = fs.glob(os.path.join(root_path, "epoch_*.pt"))
```
sagadre, line 84: is root_path here what is referred to as checkpoint_path elsewhere?

achalddave: yup!

open_lm/main.py
```diff
@@ -355,6 +362,10 @@ def main(args):
         resume_from = None
         checkpoint_path = args.checkpoint_path
 
+        # If using remote_sync, need to check the remote instead of the local checkpoints folder.
+        if args.remote_sync is not None:
+            checkpoint_path = os.path.join(args.remote_sync, args.name, "checkpoints")
```
achalddave, line 367: One question here @sagadre - we assign checkpoint_path, but we don't update args.checkpoint_path. I copied this from code from before #153. Should we just go ahead and update args.checkpoint_path too? Not sure if we use it anywhere else.

sagadre: looked over the code and this should work and not break other things. my understanding is checkpoint_path is used just to load the checkpoint, e.g., latest checkpoint from s3; however, in general we probably should simplify the checkpoint loading in a separate PR. right now it is a bit confusing with checkpoint_path, args.checkpoint_path, join(.., checkpoints), remote-sync, etc.

sagadre: maybe just having one factorized helper function to do all this loading business could be nice and also more easily unit-testable

sagadre (approved):

PR #177, opened 2023-12-20
support for local build with sagemaker

IgorVasiljevic-TRI:
Quick fix to allow debugging locally with `launch_sagemaker_train.py` with `--local` flag.


IgorVasiljevic-TRI:
@achalddave

achalddave (approved):

PR #183, opened 2023-12-26
Local tok

reinhardh:
- enabled local input paths and a test 
- added a missing line to enable gz input files
- added an option for ray dashboard in a distributed environment 
- moved the random shuffle for subfraction and and subset before taking a subset of the files, so that it'a random subfraction/subset


reinhardh:
* Added counters to keep track of the number of tokens generated by source. This is useful when using sampling to check the fractions of the tokens by source.
*  Added option to force the number of cores. The original script takes the number of cores on the machine as the number of cores available for compute, on TACC nodes this leads to resources (memory, number of threads) running out, and setting this to a lower number makes the tokenize_shuffle script usable on such nodes.

Vaishaal (approved):
this LGTM!

PR #184, opened 2023-12-28
Option to not have default_dataset_yaml added when do_sample is False and jsonl.gz option added

saurabhgarg1996:
* Option to not have default_dataset_yaml added when do_sample is False
* jsonl.gz option added


jeffreywpli (approved):
lgtm!

PR #185, opened 2023-12-31
[DRAFT] Instruction tuning

kernelmachine:
This PR adds instruction tuning to openLM. 

Currently a work in progress, but got some pretty good initial results with openLM 1B:

```
<|user|> Give me a very brief synopsis of the Donald Trump presidency.
<|assistant|>
I am not an expert on Donald Trump, but here is a brief overview of his presidency.

The inauguration of Donald Trump was historic for several reasons. First, it marked the first time that a sitting U.S. president was not re-elected in a single term. Second, it was the first time that a president was inaugurated using the "Old" or "Inaugural" Bible, which is believed to be more sacred than the "New" or "Holy" Bible. Third, it was the first time that a U.S. president was inaugurated using the "In God We Trust" motto.

In terms of policy, Donald Trump's presidency was largely focused on tax reform, regulatory rollbacks, and deregulation. In terms of foreign policy, it was largely focused on trade wars, military interventions, and military buildups.
```

### DIFF
--- a/open_lm/datapreprocess/make_assistant_data.py
+++ b/open_lm/datapreprocess/make_assistant_data.py
@@ -19,7 +19,7 @@ QUEUE_MAX = 10000
 BUFFER_MIN = 1000
 BUFFER_MAX = 200000
 CHUNK_SIZE = 2048 + 1
-SHARD_SIZE = 267
+SHARD_SIZE = 8192
 SLEEP_TIME = 1
 S3_BUCKET = "s-laion"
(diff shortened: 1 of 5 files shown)


ludwigschmidt:
Cool! Would be nice if it got the fact about presidents losing re-elections right (https://en.wikipedia.org/wiki/List_of_presidents_who_did_not_win_reelection) :-)  But still much better than the baseline without instruction tuning of course (as you said)!

PR #187, opened 2024-01-02
speeding up preprocess in tokenize_shuffle

jeffreywpli:
In `preprocess`, if a document is long (which can esp. be the case for books), then the size of `buffer` can get very large. Repeatedly slicing this large buffer to yield sequences can then be slow since it essentially involves creating a copy of the current buffer.

One fix here is to add tokens from each document into the buffer at a more controlled rate, maintaining effectively a maximum buffer size.


Vaishaal (changes requested):
Would it be possible to add a test for this somehow?

PR #188, opened 2024-01-03
fix: add contiguous to xformers_attn

ruixin31:
I encountered the following issue: 

ExecutorError: HuggingFace error: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead. Request: 
Request(model_deployment='open_lm/default-open_lm_1b', model='open_lm/default-open_lm_1b', embedding=False, prompt='Passage: The Little White Bird is a series of short episodes, including both accounts of the narrator\'s day-to-day activities ... inheritance.\nQuestion: What is bigamy?\nAnswer:', temperature=0.0, num_completions=1, top_k_per_token=1, max_tokens=100, stop_sequences=['\n'], 
echo_prompt=False, top_p=1, presence_penalty=0, frequency_penalty=0, random=None, messages=None, multimodal_prompt=None)

and I was able to fix it by adding `.contiguous()` to the output of xformer operation. 

I'm using xformers == 0.0.22.post7, and xformers==0.0.21 does not have such issue

### DIFF
--- a/open_lm/attention.py
+++ b/open_lm/attention.py
@@ -130,9 +130,10 @@ def get_attn_func(
     elif attn_name == "xformers_attn":
         return xformers_attn
     elif attn_name == "xformers_attn_variable_length":
-        # Upon changing the input sequence length, xformers changes the stride
-        # dimension of the output tensor. This makes future calls to .view()
-        # fail. One thus needs to call .contiguous() on the output tensor.
+        # Upon changing the input sequence length, xformers attention changes
+        # the stride dimension of the output tensor. This makes future calls to
+        # .view() that collapses last two dimensions fail. One thus needs to
+        # call .contiguous() on the output tensor. [#188]
         return lambda *args, **kwargs: xformers_attn(*args, **kwargs).contiguous()
     elif attn_name == "torch_attn":
         return torch_attn
--- a/open_lm/attention.py
+++ b/open_lm/attention.py
@@ -125,7 +125,9 @@ def get_attn_func(
     attn_seq_scalar=None,
     alpha=None,
 ):
-    if attn_name == "xformers_attn":
+    if attn_name == "auto":
+        return xformers_attn if torch.cuda.is_available() else torch_attn
+    elif attn_name == "xformers_attn":
         return xformers_attn
     elif attn_name == "xformers_attn_with_variable_length_support":
         return lambda *args, **kwargs: xformers_attn(*args, **kwargs).contiguous()
--- a/open_lm/attention.py
+++ b/open_lm/attention.py
@@ -129,7 +129,10 @@ def get_attn_func(
         return xformers_attn if torch.cuda.is_available() else torch_attn
     elif attn_name == "xformers_attn":
         return xformers_attn
-    elif attn_name == "xformers_attn_with_variable_length_support":
+    elif attn_name == "xformers_attn_variable_length":
+        # Upon changing the input sequence length, xformers changes the stride
+        # dimension of the output tensor. This makes future calls to .view()
+        # fail. One thus needs to call .contiguous() on the output tensor.
         return lambda *args, **kwargs: xformers_attn(*args, **kwargs).contiguous()
     elif attn_name == "torch_attn":
         return torch_attn
--- a/open_lm/attention.py
+++ b/open_lm/attention.py
@@ -34,7 +34,7 @@ def xformers_attn(queries, keys, values, is_causal):
         batch, q_seq_len, heads, _ = queries.shape
         k_seq_len = keys.shape[1]
         mask = get_rectangular_mask((batch, heads), q_seq_len, k_seq_len, queries.device, queries.dtype)
-    return xops.memory_efficient_attention(queries, keys, values, attn_bias=mask)
+    return xops.memory_efficient_attention(queries, keys, values, attn_bias=mask).contiguous()
 
 
 def torch_attn(queries, keys, values, is_causal):
--- a/open_lm/attention.py
+++ b/open_lm/attention.py
@@ -34,7 +34,7 @@ def xformers_attn(queries, keys, values, is_causal):
         batch, q_seq_len, heads, _ = queries.shape
         k_seq_len = keys.shape[1]
         mask = get_rectangular_mask((batch, heads), q_seq_len, k_seq_len, queries.device, queries.dtype)
-    return xops.memory_efficient_attention(queries, keys, values, attn_bias=mask).contiguous()
+    return xops.memory_efficient_attention(queries, keys, values, attn_bias=mask)
 
 
 def torch_attn(queries, keys, values, is_causal):
@@ -127,6 +127,8 @@ def get_attn_func(
 ):
     if attn_name == "xformers_attn":
         return xformers_attn
+    elif attn_name == "xformers_attn_with_variable_length_support":
+        return lambda *args, **kwargs: xformers_attn(*args, **kwargs).contiguous()
     elif attn_name == "torch_attn":
         return torch_attn
     elif attn_name == "custom_attn":
--- a/open_lm/params.py
+++ b/open_lm/params.py
@@ -105,8 +105,8 @@ def add_model_args(parser):
     parser.add_argument(
         "--attn-name",
         type=str,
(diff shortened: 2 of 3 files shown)


sagadre:
Is there any chance you can share a minimal script to reproduce this error?

ruixin31:
Try the `generate.py` below with the cmd line

`python scripts/generate.py --model open_lm_1b --positional-embedding-type head_rotary --checkpoint _path_to_1b_checkpoint_ --input-text 1`

Several key ingredients to trigger this bug:
- `xformers==0.0.22.post7`
- `open_lm.to(torch.bfloat16)`
- Generate >1 times with different prompts

<details>
<summary>generate.py</summary>

```python

"""Script to generate text from a trained model using HuggingFace wrappers."""

import argparse
import json
import builtins as __builtin__

import torch

from composer.utils import dist, get_device
from open_lm.utils.transformers.hf_model import OpenLMforCausalLM
from open_lm.utils.transformers.hf_config import OpenLMConfig
from open_lm.utils.llm_foundry_wrapper import SimpleComposerOpenLMCausalLM
from open_lm.model import create_params
from open_lm.params import add_model_args
from transformers import GPTNeoXTokenizerFast, LlamaTokenizerFast


builtin_print = __builtin__.print


@torch.inference_mode()
def run_model(open_lm: OpenLMforCausalLM, tokenizer, args):
    dist.initialize_dist(get_device(None), timeout=600)
    input = tokenizer(args.input_text)
    input = {k: torch.tensor(v).unsqueeze(0).cuda() for k, v in input.items()}
    composer_model = SimpleComposerOpenLMCausalLM(open_lm, tokenizer)
    composer_model = composer_model.cuda()

    generate_args = {
        "do_sample": args.temperature > 0,
        "pad_token_id": 50282,
        "max_new_tokens": args.max_gen_len,
        "use_cache": args.use_cache,
        "num_beams": args.num_beams,
    }
    # If these are set when temperature is 0, they will trigger a warning and be ignored
    if args.temperature > 0:
        generate_args["temperature"] = args.temperature
        generate_args["top_p"] = args.top_p

    output = composer_model.generate(
        input["input_ids"],
        **generate_args,
    )
    output = tokenizer.decode(output[0].cpu().numpy())
    print("-" * 50)
    print("\t\t Model output:")
    print("-" * 50)
    print(output)


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--checkpoint")
    parser.add_argument("--model", type=str, default="open_lm_1b", help="Name of the model to use")

    parser.add_argument("--input-text", required=True)
    parser.add_argument("--max-gen-len", default=200, type=int)
    parser.add_argument("--temperature", default=0.8, type=float)
    parser.add_argument("--top-p", default=0.95, type=float)
    parser.add_argument("--use-cache", default=False, action="store_true")
    parser.add_argument("--tokenizer", default="EleutherAI/gpt-neox-20b", type=str)
    parser.add_argument("--num-beams", default=1, type=int)

    add_model_args(parser)
    args = parser.parse_args()
    print("Loading model into the right classes...")
    open_lm = OpenLMforCausalLM(OpenLMConfig(create_params(args)))

    if "gpt-neox-20b" in args.tokenizer:
        tokenizer = GPTNeoXTokenizerFast.from_pretrained("EleutherAI/gpt-neox-20b")
    elif "llama" in args.tokenizer:
        tokenizer = LlamaTokenizerFast.from_pretrained(args.tokenizer)
    else:
        raise ValueError(f"Unknown tokenizer {args.tokenizer}")
    if args.checkpoint is not None:
        print("Loading checkpoint from disk...")
        checkpoint = torch.load(args.checkpoint)
        state_dict = checkpoint["state_dict"]
        state_dict = {x.replace("module.", ""): y for x, y in state_dict.items()}
        open_lm.model.load_state_dict(state_dict)
    open_lm.model.eval()

    text = """The following are multiple choice questions (with answers) about abstract algebra.

Question: Statement 1 | If aH is an element of a factor group, then |aH| divides |a|. Statement 2 | If H and K are subgroups of G then HK is a subgroup of G.
A. True, True
B. False, False
C. True, False
D. False, True
Answer: B

Question: Find all c in Z_3 such that Z_3[x]/(x^2 + c) is a field.
A. 0
B. 1
C. 2
D. 3
Answer: B

Question: Find the characteristic of the ring 2Z.
A. 0
B. 3
C. 12
D. 30
Answer: A

Question: Statement 1| Every function from a finite set onto itself must be one to one. Statement 2 | Every subgroup of an abelian group is abelian.
A. True, True
B. False, False
C. True, False
D. False, True
Answer: A

Question: Statement 1 | Every element of a group generates a cyclic subgroup of the group. Statement 2 | The symmetric group S_10 has 10 elements.
A. True, True
B. False, False
C. True, False
D. False, True
Answer: C

Question: Statement 1 | If f is a homomorphism from G to K and H is normal in G then f(H) is normal in K. Statement 2 | If f is a homomorphism from G to a group and H is finite subgroup of G, then |f(H)| divides |H|.
A. True, True
B. False, False
C. True, False
D. False, True
Answer:"""
 
    open_lm.to(torch.bfloat16)
    args.input_text = text
    run_model(open_lm, tokenizer, args)
    args.input_text += "1 "
    run_model(open_lm, tokenizer, args)




if __name__ == "__main__":
    main()
```
</details>

sagadre:
thanks for the script! will try to look into it this week. in the meantime do you think you can check the effects of this change on tokens/s/gpu (which should automatically be logged to wandb)? i am slightly concerned that this change could have a negative impact on training speed. if this is the case, it may make more sense to make changes in non-training code paths

ruixin31:
Without `.contiguous()`: 6,214.557488937922 samples_per_second_per_gpu
With `.contiguous()`: 6,215.049844771798 samples_per_second_per_gpu on a single a40 for 1 epoch

So looks like it doesn't affect performance (likely due to most data is contiguous in the first place?)

sagadre:
thanks for checking! sorry to be a stickler here, but would it be possible to check in a multinode setting? if not i will make time to check this before merging. probably everything is great, but maybe good to be careful here so that GPUs go burrr

ruixin31:
With 2 a40s: 
Without .contiguous(): 3,746.0697441510824 samples_per_second_per_gpu
With .contiguous(): 3,739.750802702916 samples_per_second_per_gpu
We observe a 0.2% decrease in speed.

I don't really know of the cause of this bug. How do I begin debugging this, since it is related to memory?

open_lm/attention.py
```diff
@@ -125,8 +125,12 @@ def get_attn_func(
     attn_seq_scalar=None,
     alpha=None,
 ):
-    if attn_name == "xformers_attn":
+    if attn_name == "auto":
+        return xformers_attn if torch.cuda.is_available() else torch_attn
+    elif attn_name == "xformers_attn":
         return xformers_attn
+    elif attn_name == "xformers_attn_with_variable_length_support":
+        return lambda *args, **kwargs: xformers_attn(*args, **kwargs).contiguous()
```
sagadre, line 133: can we add a comment here on what this code path does / when to use it e.g.,

native xformers_attention assumes that inputs are contiguous. the call to contiguous here ensures that intermediate activations remain contiguous even for variable length inputs that are not powers of 2 (e.g., sequence lengths of 51 etc.)

open_lm/params.py
```diff
@@ -105,8 +105,8 @@ def add_model_args(parser):
     parser.add_argument(
         "--attn-name",
         type=str,
-        default="xformers_attn",
-        choices=["xformers_attn", "torch_attn", "custom_attn"],
+        default="auto",
+        choices=["auto", "xformers_attn", "xformers_attn_with_variable_length_support", "torch_attn", "custom_attn"],
```
sagadre, line 109: nit: can we call this `xformers_attn_variable_length`

sagadre (approved):
left some minor comments but lgtm! thanks for debugging this! going ahead and approving so you can squash once u have addressed

ruixin31:
Note for future: 
with 
torch.Size([1, 119, 16, 128])
(243712, 2048, 128, 1)
torch.Size([1, 119, 16, 128])
(15232, 128, 15232, 1)

PR #191, opened 2024-01-04
Add shard shuffling.

GeorgiosSmyrnis:
This PR adds shuffling of shards to allow for proper testing of variance.

TODOs:

- [x] Add tests for resuming.
- [x] _Adapt previous tests._ The way the PR is written seems compatible with previous tests in which no shard shuffling is performed.


open_lm/file_utils.py
```diff
@@ -285,12 +289,15 @@ def get_string_for_epoch(
...
     if multi_epoch:
         raise NotImplementedError("Multiple passes over the dataset not fully supported yet.")
     else:
-        return _single_epoch_string(num_samples, starting_points, paths, weights, num_workers_per_gpu, world_size)
+        return _single_epoch_string(
+            num_samples, starting_points, paths, weights, num_workers_per_gpu, world_size, shard_shuffle_seed
+        )
 
 
 def _multi_epoch_string(num_samples, starting_chunk, paths, weights, min_shards_needed):
```
sagadre, line 303: just curious what this function is

sagadre: seems not used, but just want to understand a bit better

open_lm/file_utils.py
```diff
@@ -390,6 +399,15 @@ def _single_epoch_string(
     needed_samples_per_source = [int(np.ceil(weights[i] * num_samples / sum(weights))) for i in range(num_sources)]
 
     manifests = [get_metadata_file(path) for path in paths]
+    if shard_shuffle_seed is not None:
+        # Not the most efficient but probably the safest since shuffling is done inplace
+        rng_gen = np.random.default_rng(shard_shuffle_seed)
+        new_manifests = []
+        for m in manifests:
```
sagadre, line 406: i think this will break if manifests are provided in a different order in the commandline? e.g., someone passes manifest A, then manifest B vs. the other way?

sagadre: to handle this potentially re-initialize `rng_gen` within the loop?

sagadre: also do we have to call shuffle more than once (here and in `get_shards_for_chunk`)? seems safer if we can just shuffle the source shards once and then "read off" the new order of shards everywhere else?

open_lm/main.py
```diff
@@ -118,9 +129,10 @@ def load_model(args, model):
     else:
         # loading a bare (model only) checkpoint for fine-tune or evaluation
         start_epoch, global_step = 0, 0
+        pretrained_seed = None
         model.load_state_dict(checkpoint)
         logging.info(f"=> loaded checkpoint '{args.resume}' (epoch {start_epoch})")
-    return start_epoch, global_step
+    return start_epoch, global_step, pretrained_seed
```
sagadre, line 135: why do we have to return `pretrained_seed` isnt it just in `args`?

open_lm/main.py
```diff
@@ -699,7 +716,7 @@ def main(args):
         loss = CrossEntropyLossWithZLoss(args.z_loss_coefficient)
 
     if args.dataset_manifest:
-        log_num_checkpoints(total_steps, args)
+        log_num_checkpoints(total_steps, args, shard_shuffle_seed)
```
sagadre, line 719: here `shard_shuffle_seed` should be in args so probably no need to change the api?

tests/test_save_load.py
```diff
@@ -114,6 +114,94 @@ def tiny_save_load(tiny_args, fsdp=False):
...
+            epoch,
+            evaluation_metrics,
+            step=global_step,
+            is_final_checkpoint=done_training,
+            next_shard_per_source=None,
+            samples_seen=None,
+            shard_shuffle_seed=args.seed,
+        )
+
+        # Loading saved tiny model - this should fail because the seed also changes
```
sagadre, line 189: is there a way to check that the sample is correct? like if i trained 5 shards, i should start on the 6th

open_lm/file_utils.py
GeorgiosSmyrnis: It is unused for now, but it is part of a codepath for multi epoch, which will be implemented later - it currently cannot be accessed.

GeorgiosSmyrnis: I just made a wrapper between both cases here.

open_lm/file_utils.py
GeorgiosSmyrnis: Initializing rng_gen within the loop is correct, thanks for catching that!

I thought of doing the shuffling inside `get_metadata_file` (immediately after reading) but decided against it originally to preserve backwards compatibility with previous unshuffled runs. Now that I think about it again, passing it inside `get_metadata_file` is safer, as you said.

open_lm/main.py
GeorgiosSmyrnis: This is done for backwards compatibility with runs that did not have a pretrained seed - in that case, I set `pretrained_seed = None` so that no shuffling is performed (and the original behavior is preserved).

open_lm/main.py
GeorgiosSmyrnis: Same reasoning as above (but I can also just move shard_shuffle_seed inside args).

GeorgiosSmyrnis: Moved `shard_shuffle_seed` in args.

tests/test_save_load.py
GeorgiosSmyrnis: The added tests ensure determinism across both multiple calls of `get_string_for_epoch` with the same seed and across resumes.

sagadre (approved):
lgtm!

PR #192, opened 2024-01-05
refactor params

jmercat:
It seems that the Params  class is better in the params.py file than in model.py
 It allows importing params without importing the rest (which in some case might result in import loops...)
Also Params contained some functions and class members instead of containing plain parameters. I simplified this to only containing parameters. It requires creating the function or class later at usage but it's not a big overhead.
 
 This is a refactoring and might cause issues with dependencies...


Vaishaal (changes requested):
lets run a sample 160M and 1B and 7B (Cx1) with these changes to make sure nothing is breaking. This is a pretty large refactor in a critical code path. Thanks!

jmercat:
I tested it at 1B from dcnlp here:
[https://wandb.ai/tri/dcnlp/runs/jean-openlm-dcnlp-2024-01-06-00-03-36-232-llgwe9-algo-1](https://wandb.ai/tri/dcnlp/runs/jean-openlm-dcnlp-2024-01-06-00-03-36-232-llgwe9-algo-1)
and from openlm here:
[https://wandb.ai/tri/lm1/runs/jean-openlm-main-2024-01-09-23-26-32-099-iwro33-algo-1?workspace=user-jmercat](https://wandb.ai/tri/lm1/runs/jean-openlm-main-2024-01-09-23-26-32-099-iwro33-algo-1?workspace=user-jmercat)

PR #195, opened 2024-01-07
Don't double count jsonl.gz

achalddave:
Tokenize-shuffle globs jsonl.gz and .gz separately, so we end up seeing some files more than once. This fixes that.


Vaishaal (approved):
lgtm

achalddave:
Superseded by #207

PR #197, opened 2024-01-09
Enable multiple passes over the dataset.

GeorgiosSmyrnis:
This PR adds an extra flag `--multiple-data-passes` which allows for multiple passes over the dataset.
The code is set up so that the total number of tokens that will be seen across all passes of the dataset will still be `epochs * train_num_samples`, only in this case instead of erroring when data runs out, we move on to the next pass of the dataset (shuffling the shards as well, so that we see them in different order in each pass).

This is a WIP, since a few more test cases need to be added.

### DIFF
--- a/open_lm/file_utils.py
+++ b/open_lm/file_utils.py
@@ -293,55 +293,55 @@ def get_string_for_epoch(
 ):
     """See _single_epoch_string for full docstring."""
     if multi_epoch:
-        raise NotImplementedError("Multiple passes over the dataset not fully supported yet.")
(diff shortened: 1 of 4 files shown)


Vaishaal (approved):

xszheng2020:
Hi, @GeorgiosSmyrnis 

I am training a dclm 400m-4x model using a limited amount of tokens like 1B tokens allowing multiple passes.
Then I observe the training loss curve as follows. 
The shape looks strange since the training loss drops periodically. 
Could it be because we did not shuffle the shards correctly?

Any idea with this?

Best,
Xiaosen

image

PR #198, opened 2024-01-09
Make llm-foundry optional dependency

achalddave:
llm-foundry install seems to cause some issues for our ray scripts; making it optional since we only need it for eval, and added a pointer to README to install it.


achalddave:
I guess our tests rely on composer right now, so this will need to be a bit more involved. closing for now.

PR #199, opened 2024-01-10
modified eval to support downstream perplexity evals and DDP

sagadre:


open_lm/data.py
```diff
@@ -577,3 +578,87 @@ def get_data(args, epoch=0, tokenizer=None, skip_train=False, floor=True):
...
+        # create mask for positions before (last) tok in each row (batch)
+        tok_mask = cumsum_mask > 0
+
+        if args.squash_mask_left:
+            # retain tok in the output as makes squashing easier
+            tok_mask &= ~tok_positions_targets
+
+            # ensure we dont modify input
+            targets = torch.clone(targets)
+            targets[tok_mask] = ignore_tok
```
achalddave, line 627: I'm struggling a bit in this workflow. Can you add an example like the following in comments? I think it would really help. I attempted below, but I'm getting something wrong in the logic.

```# Let
# L = args.target_mask_left
# _ = ignore_tok
# P = pad token (args.target_mask_individual)
inputs                = [a, b, c, L, d, e, L, f]
targets               = [b, c, L, d, e, L, f, g]

tok_positions_targets = [0, 0, 1, 0, 0, 1, 0, 0]
cumsum_mask           = [2, 2, 2, 1, 1, 1, 0, 0]
tok_mask              = [1, 1, 1, 1, 1, 1, 0, 0]
# after tok_mask &= ~tok_positions_targets
tok_mask              = [1, 1, 0, 1, 1, 0, 0, 0]
# after targets[tok_mask] = ignore_tok; let _ = ignore_tok
targets               = [_, _, L, _, _, L, e, f]
selected_inputs       = [1, 1, 1, 0, 1, 1, 0, 1]
selected_targets      = [1, 1, 0, 1, 1, 0, 1, 1]
# after squash, let P = pad token (args.target_mask_individual)
inputs                = [a, b, c, d, e, f, P, P]
targets               = [_, _, _, _, e, f, P, P]
```

open_lm/evaluate.py
```diff
@@ -0,0 +1,191 @@
...
+    sps_m = AverageMeter()
+    spspg_m = AverageMeter()
+    losses_seq_ci_m = ConfidenceIntervalMeter()
+    losses_tok_ci_m = ConfidenceIntervalMeter()
+
+    end = time.time()
+    loss = torch.nn.CrossEntropyLoss(reduction="none")
+
+    for i, batch in enumerate(dataloader):
+        if i == max_num_batches and max_num_batches != 0:
```
achalddave, line 54: Presumably the dataloader should end exactly when this is true right? If so, I think we don't need this check. If not, we might be missing some data from eval?

tests/test_loss_masking.py
```diff
@@ -22,57 +24,25 @@ def create_tok_fixtures():
     return special_token_left, special_token_individual, batched_tokens
 
 
-def test_create_replace_before_tok():
+def test_target_mask_left():
+    args, _, _, _, _, _ = create_train_fixtures()
     special_token, _, batched_tokens = create_tok_fixtures()
```
achalddave, line 29: nit: I would actually just copy paste create_tok_fixtures here - it's duplicated but makes the tests easier to read and insepct.

open_lm/evaluate.py
sagadre: so the dataloader always tries to exhaust the data. we have a special flag `--val-num-samples` for early stoping of eval. the current convention is if max_num_batches is 0 then we exhaust the loader, but `None` makes better sense here so updating to that

sagadre: will add a comment to this effect

open_lm/data.py
sagadre: this is a great idea! admittedly the logic is kinda confusing, tried to correct the hand-simulation in place
```
# L = args.target_mask_left
# _ = ignore_tok
# P = pad token (args.target_mask_individual)
inputs                = [a, b, c, L, d, e, L, f]
targets               = [b, c, L, d, e, L, f, g]

tok_positions_targets = [0, 0, 1, 0, 0, 1, 0, 0]
cumsum_mask           = [2, 2, 2, 1, 1, 1, 0, 0]
tok_mask              = [1, 1, 1, 1, 1, 1, 0, 0]
# after tok_mask &= ~tok_positions_targets
tok_mask              = [1, 1, 0, 1, 1, 0, 0, 0]
# after targets[tok_mask] = ignore_tok; let _ = ignore_tok
targets               = [_, _, L, _, _, L, f, g] # LAST 2 ELEMENTS HERE ARE f, g, NOT e, f
selected_inputs       = [1, 1, 1, 0, 1, 1, 0, 1]
selected_targets      = [1, 1, 0, 1, 1, 0, 1, 1]
# after squash, let P = pad token (args.target_mask_individual)
inputs                = [a, b, c, d, e, f, P, P]
targets               = [_, _, _, _, f, g, P, P] # PROPOGATING HERE
```

achalddave: ahh perfect, thanks! yeah if you can add this in as comments it would help. You don't need all the steps (since the comment may go out of date when we change the code), maybe just the original inputs/targets and final inputs/targets would help

open_lm/evaluate.py
achalddave: ahh okay, makes sense, fine as is then

open_lm/data.py
```diff
@@ -577,3 +578,144 @@ def get_data(args, epoch=0, tokenizer=None, skip_train=False, floor=True):
...
+        start_idx = 0
+    elif chunk.shape[1] > args.seq_len + 1:
+        start_idx = torch.randint(0, chunk.shape[1] - args.seq_len, (1,)).item()
+    else:
+        raise Exception(f"Invalid sequence length: Sequence length {args.seq_len} > {chunk.shape[1]} Chunk size")
+
+    inputs = chunk[:, start_idx : start_idx + args.seq_len]
+    targets = chunk[:, start_idx + 1 : start_idx + args.seq_len + 1]
+
+    # replace elements to be masked with with -100 (pytorch default xent ignore value)
```
achalddave, line 717: I guess this comment is out of date right?

achalddave (approved):
LGTM, thanks! left one minor nit comment, but not blocking

PR #213, opened 2024-01-31
Attention across documents.

GeorgiosSmyrnis:
This adds a flag that stops attention from going across documents, identified by the EOT token.

The loss for the token right after the EOT token is ignored.

TODO: add some tests for the shape of the mask.


open_lm/params.py
```diff
@@ -742,6 +742,11 @@ def parse_args(args):
         action="store_true",
         help="If set, allow model to do multiple data passes over our dataset, in order to reach the desired number of tokens.",
     )
+    parser.add_argument(
+        "--mask-across-documents",
```
sagadre, line 746: i think this should be an `int` not a `bool` so that a user can specify their EOT token

open_lm/train.py
```diff
@@ -109,13 +138,21 @@ def train_one_epoch(model, data, loss, epoch, step, optimizer, scaler, scheduler
...
 
         if args.accum_freq == 1:
             with autocast():
                 inputs, targets = sample_chunk(texts, args)
-                out, _, _ = model(inputs)
+                document_seqlens = get_document_seqlens(inputs, args)
+                if args.mask_across_documents:
+                    # Some input samples contain EOT as the final token. The prediction after that is meaningless, so it
+                    # should not contribute to the loss.
+                    ignore_indices = torch.nonzero(inputs == SpecialTokens.END_OF_TEXT.value, as_tuple=True)
```
sagadre, line 152: i prefer not to hard code our EOT to keep open_lm tokenizer agnostic

open_lm/attention.py
```diff
@@ -16,58 +16,88 @@ def get_rectangular_mask(shape, q_seq_len, k_seq_len, device, dtype):
...
 
-    mask = None
-    # If queries have shape [batch, 1, heads, dim] it means there is only one query in the sequence.
-    # In this case, there is no notion of causal masking, so we can just set the mask to None.
-    # This is actually needed to get the desired behavior with seq_len=1.
-    if is_causal and queries.shape[1] == keys.shape[1]:
-        mask = xops.LowerTriangularMask()
-    elif is_causal and queries.shape[1] > 1:
-        # Build causal mask that assumes queries are in the end of the sequence.
+    if document_seqlens is None or all(len(ds) == 1 for ds in document_seqlens):
```
sagadre, line 26: is the `or ...` for performance reasons?

GeorgiosSmyrnis: Yes, pretty much - if all sequences within your batch come from the same document, then using the previous mask is much faster.

open_lm/params.py
GeorgiosSmyrnis: Makes sense - will update the parameter.

open_lm/train.py
GeorgiosSmyrnis: Agreed - I'll change it so that it uses the user defined EOT token.

open_lm/train.py
```diff
@@ -135,6 +172,12 @@ def train_one_epoch(model, data, loss, epoch, step, optimizer, scaler, scheduler
             per_batch = args.per_gpu_batch_size // args.accum_freq
 
             inputs, targets = sample_chunk(texts, args)
+            if args.mask_across_documents:
+                # Some input samples contain EOT as the final token. The prediction after that is meaningless, so it
+                # should not contribute to the loss.
+                ignore_indices = torch.nonzero(inputs == SpecialTokens.END_OF_TEXT.value, as_tuple=True)
+                targets = targets.detach().clone()  # Clone this because it shares mem with input!
```
achalddave, line 179: Interesting, is the detach necessary here? When args.mask_across_documents is False, should we also a detach()?

GeorgiosSmyrnis: Detach is not necessary, but clone is - because the targets and the input share the underlying tensor, if the target is explicitly set then the input is also affected.

When args.mask_across_documents is False, this is not an issue - neither the target nor the input are explicitly changed.

PR #214, opened 2024-02-01
allow loading with a different seed with an input flag

jmercat:
I think this check should actually not be done in the loading function since it has to do with training not loading....
Here for example, I want to use this loading function for eval and I don't care about the seed for shard shuffling. To make this quicker I added a flag but a refactor is needed later.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -100,18 +100,20 @@ def get_state_dict(name):
     return sd
 
 
-def load_model(args, model):
+def load_model(args, model, different_seed=False):
     checkpoint = pt_load(args.resume, map_location="cpu")
     if "epoch" in checkpoint:
-        if "shard_shuffle_seed" in checkpoint:
+        if not different_seed and "shard_shuffle_seed" in checkpoint:
             pretrained_seed = checkpoint["shard_shuffle_seed"]
             assert (
                 pretrained_seed == args.seed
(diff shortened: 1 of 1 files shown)


open_lm/main.py
```diff
@@ -100,18 +100,20 @@ def get_state_dict(name):
...
         else:
-            logging.info(
-                "Resuming a checkpoint that does not have a seed saved. This means that the shards were not shuffled, so they will remain unshuffled."
-            )
+            if "shard_shuffle_seed" in checkpoint:
+                message = "Resuming a checkpoint that does not have a seed saved. This means that the shards were not shuffled, so they will remain unshuffled."
+            else:
+                message = "Resuming a checkpoint without checking that the seed match. This means that training might not be reproducible."
+            logging.info(message)
             pretrained_seed = None
```
GeorgiosSmyrnis, line 117: I think the structure should be better as e.g. if different_seed then print a message that we ignore seed checking
and if not different_seed then do the exact same thing as before. I think this makes the messages more consistent.

GeorgiosSmyrnis (commented):
Maybe a good way to do this (if you want to eval with different seeds) is to check if there is training data and infer if the run is for training or not (instead of using a different flag)?

GeorgiosSmyrnis (approved):
Approving as this is only for messages.

PR #218, opened 2024-02-21
Adding support for FP8 training

rams16592:
- Changes made to `train.py`, `model.py`, and `norms.py` to check if Transformer Engine can be imported for P5 instances or H100s and use FP8 for Linear and LayerNorm layers. 
- Minor modifications to `main.py` for FP8 support

### DIFF
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -35,6 +35,17 @@ try:  # optional import
 except ImportError:
     MambaLMHeadModel = None
 
+# Adding flag if using TE FP8
+using_te = False
+try:
+    import transformer_engine.pytorch as te
+    from transformer_engine.common import recipe
+    fp8_format = recipe.Format.HYBRID
+    fp8_recipe = recipe.DelayedScaling(fp8_format=fp8_format, amax_history_len=32, amax_compute_algo="max")
+    using_te = True
+except ImportError as ie:
+    using_te = False
+
 # from openclip
 _MODEL_CONFIG_PATHS = [Path(__file__).parent / f"model_configs/"]
 _MODEL_CONFIGS = {}  # directory (model_name: config) of model architecture configs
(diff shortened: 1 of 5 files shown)


open_lm/model.py
```diff
@@ -117,41 +128,72 @@ def __init__(self, layer_id, args: Params):
         super().__init__()
         self.n_heads = args.n_heads
         self.head_dim = args.dim // args.n_heads
-        self.in_proj = nn.Linear(args.dim, 3 * args.n_heads * self.head_dim, bias=False)
-        self.out_proj = nn.Linear(args.n_heads * self.head_dim, args.dim, bias=False)
+        if using_te:
```
sagadre, line 131: instead of if/else's can we have a single helper function that recursively searches the model and replaces the linears?

open_lm/model.py
```diff
@@ -117,41 +128,72 @@ def __init__(self, layer_id, args: Params):
...
-        # scale init by depth as in https://arxiv.org/abs/1908.11365 -- worked slightly better.
-        std = std / math.sqrt(2 * (self.layer_id + 1))
-        torch.nn.init.trunc_normal_(self.out_proj.weight, std=std, a=-3 * std, b=3 * std)
+        if using_te:
+            # initialize weights by trunc_normal(1/sqrt(fan_in))
+            std = 1.0 / math.sqrt(self.dim)
+            torch.nn.init.trunc_normal_(self.in_proj.weight_tensor.float(), std=std, a=-3 * std, b=3 * std)
+            # scale init by depth as in https://arxiv.org/abs/1908.11365 -- worked slightly better.
+            std = std / math.sqrt(2 * (self.layer_id + 1))
+            torch.nn.init.trunc_normal_(self.out_proj.weight_tensor.float(), std=std, a=-3 * std, b=3 * std)
```
sagadre, line 189: why do we need to cast to float? does a float cast happen in place?

open_lm/model.py
sagadre: another option is having a linear/layernorm module `m` that is set to either `te` or `nn`

open_lm/model.py
```diff
@@ -222,14 +270,24 @@ def __init__(self, layer_id, args: Params):
...
+            self.attention_norm = te.LayerNorm(
+                args.dim,
+                eps=args.norm_eps,
+            )
+            self.ffn_norm = te.LayerNorm(
+                args.dim,
+                eps=args.norm_eps,
+            )
+        else:
+            self.attention_norm = args.norm_type(
```
sagadre, line 283: can we just add te.LayerNorm as one of the `args.norm_type`?

open_lm/norms.py
```diff
@@ -55,7 +67,16 @@ def reset_parameters(self) -> None:
                 self.bias.zero_()
 
     def forward(self, input: Tensor) -> Tensor:
-        return F.layer_norm(input, self.normalized_shape, self.weight, self.bias, self.eps)
+        if using_te:
```
sagadre, line 70: @achalddave do u think we should have a seperate class for TeLayerNorm? or do u prefer combining it with existing layer norm

open_lm/params.py
```diff
@@ -742,6 +742,17 @@ def parse_args(args):
...
         help="If set, allow model to do multiple data passes over our dataset, in order to reach the desired number of tokens.",
     )
+    parser.add_argument(
+        "--use-smp-flash-attention",
+        type=int,
+        default=None,
+        help="Using SMP Flash Attention.",
+    )
+    parser.add_argument(
+        "--sharding-strategy",
```
sagadre, line 752: is this used? if so can we have a more specific name?

sagadre: also not seeing where `--use-smp-flash-attention` is used

rams16592: This is not used for FP8. Just placeholder flags defaulted to `None` for Sagemaker Model Parallel.

open_lm/model.py
rams16592: We dont need float cast. Removing that in next commit.

open_lm/model.py
rams16592: Ideally, yes. But that would replace the Linear layer found [here](https://github.com/mlfoundations/open_lm/blob/main/open_lm/model.py#L305) which breaks the training. So until a fix is not found, need to isolate that particular layer.

open_lm/model.py
```diff
@@ -202,9 +245,14 @@ def __init__(self, layer_id, args: Params):
         elif args.ffn_type == "gelu":
```
achalddave, line 245: Could we also support fp8 for swiglu above? We can make a copy of the Swiglu class in this file. Here's the source for Swiglu https://github.com/facebookresearch/xformers/blob/7f8c290183344343771f4e1d945a8ce10a9500ff/xformers/ops/swiglu_op.py#L430

open_lm/model.py
sagadre: can't we just not recurse for special cases?

achalddave: missed this earlier, but yeah this might also address my swiglu comment below (since it would recurse and replace the linear within the swiglu)

open_lm/model.py
sagadre: @rams16592 seems like the recursive replace linear patten should take care of this automatically. a function like this seems like it would be great and we can exclude certain linears that need to be higher precision for stability. this function has an include field instead of exclude, but hopefully that's easy to flip: 
https://github.com/mlfoundations/open_clip/blob/73fa7f03a33da53653f61841eb6d69aef161e521/src/open_clip/utils.py#L65

romilshahtri: Applied this change in the latest commit. Excluding the last output Linear layer from the conversion to TE Linear as its running into errors.

rams16592: Applied this change in the latest commit. Excluding the last output Linear layer from the conversion to TE Linear as its running into errors.

open_lm/params.py
rams16592: Removed this to avoid confusion

open_lm/norms.py
rams16592: We can have a separate class for TeLayerNorm.

open_lm/model.py
rams16592: Added this in the latest commit based on presence of TE, TE.LayerNorm or NN.LayerNorm will be considered.

open_lm/model.py
rams16592: Removed completely as we are recursively changing NN.Linear to TE.Linear

open_lm/model.py
rams16592: Completed in the latest commit.

PR #223, opened 2024-03-01
added InContextLearningCodeEvalAccuracy to evaluation metrics

mayeechen:
human_eval tasks have a corresponding metric type, InContextLearningCodeEvalAccuracy, which was not included in the llm foundry wrapper. As a result, human_eval metrics were not being written to the evaluation output file.


Vaishaal (approved):
LGTM

PR #224, opened 2024-03-02
Add support for GeGLU

achalddave:
Similar to Gemma but with an expansion ratio that matches our Swiglu

### DIFF
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -242,8 +242,8 @@ class Block(nn.Module):
             self._ff_w2 = nn.Linear(self.hidden_dim, args.dim, bias=False)
             self.feed_forward = nn.Sequential(self._ff_w1, nn.GELU(approximate="none"), self._ff_w2)
         elif args.ffn_type == "gemma_geglu":
-            # Follows https://github.com/google/gemma_pytorch/blob/01062c9ef4cf89ac0c985b25a734164ede017d0b/gemma/model.py#L182-L201
-            self.hidden_dim = args.dim * 8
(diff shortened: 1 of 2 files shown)


open_lm/model.py
```diff
@@ -207,6 +241,10 @@ def __init__(self, layer_id, args: Params):
             self._ff_w1 = nn.Linear(args.dim, self.hidden_dim, bias=False)
             self._ff_w2 = nn.Linear(self.hidden_dim, args.dim, bias=False)
             self.feed_forward = nn.Sequential(self._ff_w1, nn.GELU(approximate="none"), self._ff_w2)
+        elif args.ffn_type == "gemma_geglu":
+            # Follows https://github.com/google/gemma_pytorch/blob/01062c9ef4cf89ac0c985b25a734164ede017d0b/gemma/model.py#L182-L201
+            self.hidden_dim = args.dim * 8
```
GeorgiosSmyrnis, line 246: I might have missed something, but where does the 8 come from?

GeorgiosSmyrnis (approved):
LGTM overall (just one small clarification)

open_lm/model.py
achalddave: d'oh, good catch, looks like I forgot to commit the part that matches the hidden_dim to that of swiglu. fixing...

sedrick-keh-tri:
https://twitter.com/danielhanchen/status/1763613620909580505
https://unsloth.ai/blog/gemma-bugs

Might be relevant. Seems like the original gemma uses approximate gelu instead of exact gelu

PR #234, opened 2024-03-13
main.py: set_grad_checkpointing before DDP or FSDP wrap

iejMac:
DDP + --grad-checkpointing resulted in but works with FSDP

```
AttributeErrorAttributeErrorAttributeError: : 'DistributedDataParallel' object has no attribute 'set_grad_checkpointing
': 'DistributedDataParallel' object has no attribute 'set_grad_checkpointing''DistributedDataParallel' object has no at
tribute 'set_grad_checkpointing'    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")  
AttributeError: 'DistributedDataParallel' object has no attribute 'set_grad_checkpointing' 
```

I guess FSDP passes along set_grad_checkpointing to the module but not DDP. This works with both configurations

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -439,6 +439,9 @@ def main(args):
 
     random_seed(args.seed, args.rank)
 
+    if args.grad_checkpointing:
+        model.set_grad_checkpointing()
+
     if args.distributed:
         if args.fsdp:
             transformer_layer_cls = None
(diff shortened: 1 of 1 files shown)


achalddave:
looks great, thanks @iejMac!

achalddave (approved):

PR #235, opened 2024-03-15
Generation fixes

achalddave:
Results in 18.7 TriviaQA accuracy for a 1B model, compared to ~0% earlier due to the lack of attention_mask support.

### DIFF
--- a/open_lm/attention.py
+++ b/open_lm/attention.py
@@ -6,35 +6,65 @@ import xformers.ops as xops
 
 
 def get_rectangular_mask(shape, q_seq_len, k_seq_len, device, dtype):
+    """
+    >>> get_rectangular_mask((1, 1), 2, 2, "cpu", torch.float32)
+    tensor([[[[0., -inf],
+              [0., 0.]]]])
+    >>> get_rectangular_mask((1, 1), 3, 5, "cpu", torch.float32)
+    tensor([[[[0., 0., 0., -inf, -inf],
+              [0., 0., 0., 0., -inf],
+              [0., 0., 0., 0., 0.]]]])
+    >>> get_rectangular_mask((1, 1), 5, 5, "cpu", torch.float32)
+    tensor([[[[0., -inf, -inf, -inf, -inf],
+              [0., 0., -inf, -inf, -inf],
+              [0., 0., 0., -inf, -inf],
+              [0., 0., 0., 0., -inf],
+              [0., 0., 0., 0., 0.]]]])
+    """
     # xformers requires the mask to be built with a shape that is a multiple of 8
     # probably because of the way it is implemented in CUDA
     next_multiple_8 = (k_seq_len + 7) // 8 * 8  #
-    mask = torch.ones((q_seq_len, next_multiple_8), device=device, dtype=bool)
+
+    mask = torch.ones((q_seq_len, k_seq_len), device=device, dtype=bool)
     mask[:, -q_seq_len:] = torch.tril(mask[:, -q_seq_len:], diagonal=0)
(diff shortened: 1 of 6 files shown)


open_lm/attention.py
```diff
@@ -6,35 +6,65 @@
...
+
+    mask = torch.ones((q_seq_len, k_seq_len), device=device, dtype=bool)
     mask[:, -q_seq_len:] = torch.tril(mask[:, -q_seq_len:], diagonal=0)
-    return torch.zeros((*shape, q_seq_len, next_multiple_8), device=device, dtype=dtype).masked_fill(
-        ~mask, float("-inf")
-    )[:, :, :, :k_seq_len]
+
+    output_mask = torch.zeros((*shape, q_seq_len, next_multiple_8), device=device, dtype=dtype)
+    output_mask[:, :, :, :k_seq_len].masked_fill_(~mask, torch.finfo(dtype).min)
+    return output_mask[:, :, :, :k_seq_len]
```
jmercat, line 33: I thought you needed to select -k_seq_len: here

open_lm/attention.py
```diff
@@ -6,35 +6,65 @@
...
+    elif is_causal and (queries.shape[1] > 1 or attention_mask is not None):
         # Build causal mask that assumes queries are in the end of the sequence.
         batch, q_seq_len, heads, _ = queries.shape
         k_seq_len = keys.shape[1]
-        mask = get_rectangular_mask((batch, heads), q_seq_len, k_seq_len, queries.device, queries.dtype)
-    return xops.memory_efficient_attention(queries, keys, values, attn_bias=mask)
+        bias = get_rectangular_mask((batch, heads), q_seq_len, k_seq_len, queries.device, queries.dtype)
+        if attention_mask is not None:
+            # Update mask to remove attention based on attention_mask that's passed in.
+            assert attention_mask.dim() == 2
```
jmercat, line 56: the attention mask has to be the same for each element of the batch?

open_lm/model.py
```diff
@@ -290,12 +291,14 @@ def reset_parameters(self):
             std = std / math.sqrt(2 * (self._layer_id + 1))
             torch.nn.init.trunc_normal_(self._ff_w2.weight, std=std, a=-3 * std, b=3 * std)
 
-    def forward(self, x, past_key_value=None, use_cache=False):
+    def forward(self, x, past_key_value=None, use_cache=False, attention_mask=None):
+        # import ipdb; ipdb.set_trace()
```
jmercat, line 295: could remove

tests/test_attention_masking.py
```diff
@@ -0,0 +1,31 @@
...
+    output1 = xformers_attn(
+        queries[:, 4:],
+        keys[:, 4:],
+        values[:, 4:],
+        is_causal=True,
+        attention_mask=attention_mask[:, 4:],
+    )
+
+    # Run with all elements but mask the first 4 elements
+    output2 = xformers_attn(queries, keys, values, is_causal=True, attention_mask=attention_mask)
```
jmercat, line 28: should it also test that when the given attention mask is just a causal mask then the output is the same as not passing a mask at all?

open_lm/attention.py
achalddave: no no, the attention_mask huggingface provides is actually (num_sequences, sequence_length), and is set to False for the indices that should not be attended to.

open_lm/attention.py
achalddave: I fixed it above instead by setting up the first few columns of output_mask appropriately (instead of the last few columns)

open_lm/model.py
achalddave: d'oh thanks will do

jmercat (approved):

achalddave:
thanks @jmercat!

PR #236, opened 2024-03-20
v2 small-scale configs

sagadre:
small-scale configs from language models scale reliably with over-training and on downstream tasks (https://arxiv.org/abs/2403.08540)


GeorgiosSmyrnis (approved):
LGTM - should we consider removing the old versions as well?

sagadre:
its a good question. i think probably we should keep old configs in case others are using them. for example, overriding `open_lm_11m` may cause issues for some people

i think lets stick with *_v2 for now

PR #238, opened 2024-03-24
Adding averaging of iterates

Tomerporian:
Supports EMA and polynomial averaging. Includes option for more than one averaged model and taking intermittent averaging.
For example, running EMA with parameter 0.999 and period 100 and polynomial averaging with parameter 8 and period 10, use
`--averagers="ema_0.999_100,poly_8_10"`
Logging of training loss of the averaged models can be done by choosing `--log-avg-model-training-loss` to be bigger than zero.

### DIFF
--- a/open_lm/main.py
+++ b/open_lm/main.py
@@ -32,7 +32,7 @@ from torch.distributed.fsdp.wrap import transformer_auto_wrap_policy
 from open_lm.data import proc_token
 from open_lm.model import Block
 from open_lm.losses import CrossEntropyLossWithZLoss
-
+from open_lm.utils.average_utils import ModelAverager
 try:
     import wandb
 except ImportError:
@@ -50,7 +50,7 @@ from open_lm.data import get_data, get_wds_dataset
 from open_lm.distributed import is_master, init_distributed_device, broadcast_object
 from open_lm.logger import setup_logging
 from open_lm.params import parse_args
-from open_lm.scheduler import cosine_lr
+from open_lm.scheduler import cosine_lr, const_lr
 from open_lm.train import train_one_epoch
 from open_lm.evaluate import evaluate_loop
 from open_lm.file_utils import (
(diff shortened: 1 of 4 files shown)


open_lm/main.py
```diff
@@ -636,14 +668,27 @@ def main(args):
...
+            scheduler = const_lr(
+                optimizer,
+                args.lr,
+                args.warmup,
+                # total_steps,
+                # args.lr_cooldown_end,
+                # args.force_min_lr,
+            )
+        else:
+            logging.error(f"Unknown scheduler, {args.lr_scheduler}. Available options are: cosine, const.")
```
achalddave, line 690: Can you make this raise a ValueError exception instead?

open_lm/params.py
```diff
@@ -237,7 +237,7 @@ def check_args(args):
             if args.remote_sync_protocol != "s3":
                 raise ValueError("Sync protocol not supported when using resume latest.")
 
-    if args.lr_scheduler != "cosine":
+    if args.lr_scheduler != "cosine" or args.lr_scheduler != "const" or args.lr_scheduler != "const-cooldown":
```
achalddave, line 240: nit: Can you replace with the below for readability?
```if args.lr_scheduler not in {"cosine", "const", "const-cooldown"}:```

achalddave (commented):
Overall this looks great, thanks @Tomerporian! Left a few small comments. I'll run a 1B model with a few settings for averagers, and assuming that goes smoothly, we should be good to merge!

open_lm/params.py
achalddave: actually the current version also has a bug which causes the unit tests to fail. more generally, maybe we can move the lr_scheduler components out of this PR?

open_lm/main.py
Tomerporian: sure no problem

open_lm/params.py
Tomerporian: i added the line as you wrote

achalddave (approved):

PR #242, opened 2024-04-06
Presorting tokenize shuffle (useful for in-context-learning)

revbucket:
Added a presort flag to tokenize shuffle that we need for in-context-pretraining. 

We can remove this flag at any point, or the ICLM tok/shuffle steps can be done from this branch. Either/or


achalddave:
can you fix the lint @revbucket? then we should merge

achalddave (approved):

PR #246, opened 2024-04-11
Input embed

jmercat:
This allows calling the model with inputs_embeds arguments instead of input_ids if needed.

It also removes llm-foundry from the requirements. It is already explained in the README that llm-foundry is a separate additional installation to make. 
llm-foundry is extremely strict in its requirements and makes a lot of packages incompatible with OpenLM so I don't think it should be required. Moreover it's only used for eval so it is not actually required by the main model.

### DIFF
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -396,7 +396,7 @@ class Transformer(nn.Module, PyTorchModelHubMixin):
             x = inputs_embeds
         else:
             raise ValueError("Either input_ids or inputs_embeds must be provided.")
-        
+
         x = self.post_embed_norm(x)
 
         if past_key_values is None:
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -380,7 +380,7 @@ class Transformer(nn.Module, PyTorchModelHubMixin):
     def set_grad_checkpointing(self, enable=True):
         self.grad_checkpointing = enable
 
-    def forward(self, input, past_key_values=None, use_cache=False, attention_mask=None):
+    def forward(self, input_ids=None, inputs_embeds=None, past_key_values=None, use_cache=False, attention_mask=None):
         """
         Args:
             input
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -24,4 +24,4 @@ transformers
 s3fs
 wikipedia
 ipython
-llm-foundry>=0.4.0
+# llm-foundry>=0.4.0
```
sagadre, line 27: agree its ok to remove this as a dependency. can we guard the `import` here with a try/catch telling the user to pip install or pointing to the README? : https://github.com/mlfoundations/open_lm/blob/6bb1247d8e668099fdfcacbe6a42d4f3ba87d5cc/eval/eval_openlm_ckpt.py#L10

open_lm/utils/llm_foundry_wrapper.py
```diff
@@ -50,5 +50,5 @@ def __init__(self, model, tokenizer):
             shift_labels=True,
         )
 
-    def generate(self, input_ids, **kwargs):
+    def generate(self, input_ids=None, **kwargs):
```
sagadre, line 53: maybe add `input_embeds` explicitly so the api is more transparent?

open_lm/model.py
```diff
@@ -380,7 +380,7 @@ def reset_parameters(self):
     def set_grad_checkpointing(self, enable=True):
         self.grad_checkpointing = enable
 
-    def forward(self, input, past_key_values=None, use_cache=False, attention_mask=None):
+    def forward(self, input_ids=None, inputs_embeds=None, past_key_values=None, use_cache=False, attention_mask=None):
```
sagadre, line 383: nice!

sagadre (approved):
left some nits, but overall lgtm. feel free to merge before/after addressing nits

sagadre:
oh also lets get a version bump here for downstream project pip installs:
https://github.com/mlfoundations/open_lm/blob/6bb1247d8e668099fdfcacbe6a42d4f3ba87d5cc/setup.py#L22

achalddave:
this is great, one request - can you add llm-foundry to requirements-test, or add a pytest decorator around tests that require llm-foundry? currently some tests are failing because they rely on llm-foundry

open_lm/utils/llm_foundry_wrapper.py
jmercat: Adding input_embeds created a bug in composer.... I grew older while finding it came from that T-T

PR #250, opened 2024-04-15
Add attention masking support for torch_attn

achalddave:


open_lm/attention.py
```diff
@@ -33,6 +36,33 @@ def get_rectangular_mask(shape, q_seq_len, k_seq_len, device, dtype):
...
+    Args:
+        bias (torch.Tensor, shape (batch_size, num_heads, q_seq_len, k_seq_len))
+        attention_mask (torch.Tensor, shape (batch_size, sequence_len))
+        queries_dtype: queries.dtype; used to get minimum value for masked indices.
+
+    Returns:
+        bias_with_mask (torch.Tensor, shape (batch_size, num_heads, q_seq_len, k_seq_len))
+    """
+    # Update mask to remove attention based on attention_mask that's passed in.
+    assert attention_mask.dim() == 2
```
GeorgiosSmyrnis, line 51: This forces the attention mask to be the same across all samples - while I understand the reasoning with this limitation (xformers is kind of weird about this I think), it might be worth adding an error message.

tests/test_attention_masking.py
```diff
@@ -43,3 +43,44 @@ def test_attention_masking1():
 
     assert torch.allclose(output_no_mask, output_dummy_mask)
     assert torch.allclose(output_no_mask, output_mask_initial[:, 4:])
+
+
+@pytest.mark.gpu
```
GeorgiosSmyrnis, line 48: Is it that slow without gpu? This seems like a test that will be good to run automatically.

GeorgiosSmyrnis (approved):
Approved with some minor comments.

PR #252, opened 2024-04-16
Add xformers vs torch test

achalddave:


GeorgiosSmyrnis:
Looks good, I think the only other test that would be good is the `nquery < nkeys` test as you mentioned offline.

GeorgiosSmyrnis (approved):
LGTM - thanks a lot @achalddave !

PR #254, opened 2024-04-18
Mamba update

jmercat:
Mamba is now using a class for its input, this updates OpenLM accordingly.


arjunsesh:
High level comment that these changes are up to date as of mamba-ssm 1.2.1 where I have been able to successfully run the code (however gradient_checkpointing does not work). There are now more features with mamba-ssm>=2.0.0, and submodules have been renamed, so there are some changes to be made to make it compatible.

jmercat:
I had the same problem with gradient checkpointing and it prevented some interesting experiments about long-context mamba... I didn’t have time to look into this yet but maybe updating to Mamba2 first would make sense.

PR #263, opened 2024-04-30
Parameter input rotary-freq

jmercat:
This allows to change the rotary positional embedding frequency parameter.
This is useful given the more recent approaches:
LLaMA 1&2 used 10000 which is the default value here.
LLaMA 3 uses 500000, Mistral uses 1000000
LLaMA long context extension works usually increase from 10000 to 100000.

### DIFF
--- a/open_lm/model.py
+++ b/open_lm/model.py
@@ -98,15 +98,16 @@ class Params:
     moe_top_k: int = 2
     moe_freq: int = 0
     positional_embedding_type: str = "rotary"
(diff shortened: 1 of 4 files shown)


jmercat:
To test my changes I ran training with a small model and put a breakpoint where the parameter is used in rotary.py. I set a different value both in command line or in the json file and it was correctly passed on. Training was successful and tests passed. (except one in grad_accum and one in tokenize but not due to this PR, just access right issues).

image

PR #266, opened 2024-05-01
use different_seed=True when args.load_pretrained_state is false

jeffreywpli:
Allows for using a different seed when starting from a pretrained checkpoint but not loading the pretrained state.


GeorgiosSmyrnis (approved):
LGTM!

PR #268, opened 2024-05-02
Don't set log level for downstream modules

### LINKED ISSUE #261: Reduce logging when --torchcompile is passed
Currently, when we compile our models, there is a slew of logging messages from torch that make it hard to find potential errors and to read training logs reasonably. We should figure out how to disable these; likely the root cause is that we set the log level for the root logger to INFO in https://github.com/mlfoundations/open_lm/blob/main/open_lm/logger.py, and we should instead set the log level to WARN there, then use an openlm specific logger that's set to log at INFO.

achalddave:
Closes #261

### DIFF
--- a/open_lm/logger.py
+++ b/open_lm/logger.py
@@ -14,9 +14,6 @@ def setup_logging(log_file, level, include_host=False):
         formatter = logging.Formatter("%(asctime)s | %(levelname)s | %(message)s", datefmt="%Y-%m-%d,%H:%M:%S")
 
     logging.root.setLevel(level)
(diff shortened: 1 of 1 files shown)


GeorgiosSmyrnis (approved):
LGTM - does this suppress other logs apart from the ones mentioned in #261 ?

achalddave:
This will suppress all the info logs from modules outside openlm. Right now all our openlm logs are on the root logger, so those still pass through. I think this is the right thing to do - we don't need INFO logs from all modules. But I could also specifically skip the torch modules if you or anyone else feels strongly (@Vaishaal @sagadre)

PR #271, opened 2024-05-10
make tokenize-shuffle more robust to long pages

jeffreywpli:
Current Issue: 

Some sources (e.g. code repos, books) are prone to having very long pages leading to long lists of tokens. However, the way we currently yield sequences is to simply add all tokens from a page to the buffer and to then repeatedly slice the buffer. When buffer grows too large this can be very slow (and many slices need to take place). 

Solution:

Control the max buffer size by yielding tokens from a page's token set to the buffer in a more controlled manner.


Vaishaal (approved):
LGTM

PR #272, opened 2024-05-11
Even better fix for long pages

jeffreywpli:


GeorgiosSmyrnis (commented):
Just for me to understand the intent of this, is the goal to not decrease the buffer every iteration, but instead only at the very end (when you run out of tokens)?

GeorgiosSmyrnis (approved):
Approved per offline discussion - this greatly improves runtime

PR #280, opened 2024-05-18
Add option to predownload data from s3 at the start of each checkpoint.

GeorgiosSmyrnis:
This adds an option that allows for predownloading of data at the start of each checkpoint to local storage. This helps with potential s3 errors.


open_lm/file_utils.py
```diff
@@ -481,3 +483,39 @@ def _single_epoch_string(
...
+            for sf in shard_ids:
+                aws_command.extend(["--include", f"{sf}.tar"])
+
+            result = subprocess.run(
+                aws_command,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
+            if result.returncode != 0:
+                raise RuntimeError(f"Error: Failed to download data to local storage: {result.stderr.decode('utf-8')}")
```
achalddave, line 516: Should we try this a few times before erroring?

achalddave: do we need a barrier after the copy so the other workers don't start reading before worker 0 finishes copying?

GeorgiosSmyrnis: I'll add retries - there is already a dist barrier to make sure the training string is the same across all nodes, so no need to add more.

PR #281, opened 2024-05-20
Add more detailed timing.

GeorgiosSmyrnis:
This PR adds more detailed timing for our training run, so that performance can be monitored more closely.


GeorgiosSmyrnis:
This now fixes an issue where the time to sync loss across ranks was not included in the batch time.

achalddave (approved):

PR #286, opened 2024-06-12
fix: adopt latest state_dict processing

ruixin31:
fix https://github.com/mlfoundations/open_lm/commit/87309a4488a9b4c52408e2957cb3c515b2beedd9 induced model loading issue


achalddave (approved):

achalddave:
approved but I think we need to address #287 first so tests pass, we can merge this after

PR #287, opened 2024-06-25
use llm-foundry for ICL metrics

ysharma1126:
composer [deprecated](https://github.com/mosaicml/composer/blob/dev/composer/metrics/nlp.py#L184) support for ICL metrics, should now be using llm-foundry instead


achalddave:
thanks for the PR! Any reason for removing the InContextLearningQAAccuracy and InContextLearningCodeEvalAccuracy in this PR?

ysharma1126:
oh those deprecated metrics aren't in foundry (not sure why?), I guess those should be directly added to the foundry wrapper, unless ofc composer is simply fixed to the version prior to deprecation (can do it myself if you lmk the version!)

ysharma1126:
updated!

achalddave (approved):

PR #289, opened 2024-06-30
Relax boto requirement

achalddave:
not sure why we tied it to this old version, causes some requirements issues with other packages that rely on openlm


achalddave:
oops, duplicate of #284

